PR #15, opened 2025-09-17
Initial implementation of the strata predicate format

prajwolrg:
### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

### DIFF
--- a/crates/predicate/src/lib.rs
+++ b/crates/predicate/src/lib.rs
@@ -97,7 +97,6 @@ pub use type_ids::PredicateTypeId;
 
 // Internal imports for the verify_claim_witness function
 use errors::PredicateResult;
-use verifiers::VerifierType;
 
 /// Verifies that a witness satisfies a predicate key for a given claim.
 ///
@@ -117,6 +116,5 @@ pub fn verify_claim_witness(
     witness: &[u8],
 ) -> PredicateResult<()> {
     let predicate = PredicateKeyBuf::try_from(predicate_bytes)?;
-    let verifier = VerifierType::from(predicate.id());
-    verifier.verify_claim_witness(predicate.condition(), claim, witness)
+    predicate.verify_claim_witness(claim, witness)
 }


crates/predicate/Cargo.toml
storopoli, line 1: From the STR-1674:

> The crate should have the following feature flags, in order to ensure that irrelevant signing algorithms doesn’t end up compiled into binaries like the proofs that don’t need it:
> 
> - `verify` (enabled by default)
> 
> - `gen-schnorr` (disabled by default)
> 
> - `gen-sp1` (disabled by default, maybe actually ends up being more work than useful to have)
> 
> - whatever other predicate types are implemented

crates/predicate/src/verifiers/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+///
+/// This enum serves as a dispatch mechanism to route verification calls to the
+/// appropriate predicate-specific implementation based on the predicate type.
+/// Each variant contains the implementation for a specific predicate backend.
+#[derive(Debug)]
+pub(crate) enum PredicateImpl {
+    /// Never accept predicate implementation (always fails verification).
+    NeverAccept,
+    /// Always accept predicate implementation (for testing and placeholders).
+    AlwaysAccept,
```
storopoli, line 29: This is dangerous, should be featured flag under `#[cfg(test)]`

crates/predicate/src/verifier.rs
```diff
@@ -0,0 +1,47 @@
...
+//! The main verification entry point [`verify_claim_witness`] is located in the crate root.
+
+use crate::errors::Result;
+
+/// Base trait for implementing predicate type-specific verification logic.
+///
+/// This trait defines the interface for predicate implementations as specified in
+/// SPS-predicate-fmt. Each predicate type (Always Accept, BIP-340 Schnorr, SP1 Groth16, etc.)
+/// implements this trait to provide parsing and verification capabilities.
+pub(crate) trait PredicateVerifier {
```
storopoli, line 16: shouldn't this trait be `pub` and not `pub(crate)` and re-exported from the `lib.rs`?

crates/predicate/src/key.rs
```diff
@@ -0,0 +1,149 @@
...
+///
+/// Each predicate type has its own interpretation of the condition data:
+///
+/// An empty predicate key represents the "never accept" predicate.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct PredicateKey {
+    /// Raw bytes containing the type (first byte) and condition data (remaining bytes).
+    /// If empty, represents the "never accept" predicate type.
+    data: Vec<u8>,
+}
```
storopoli, line 20: Given the docstring I think we should break `data` into `type` and `condition`. And create a `TryFrom` for `&[u8]`.

crates/predicate/src/key.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// The format is: [type: u8][condition: bytes...]
+    /// An empty byte array represents the "never accept" predicate.
+    ///
+    /// # Arguments
+    /// * `bytes` - Serialized predicate key data
+    pub fn from_bytes(bytes: &[u8]) -> Self {
+        Self {
+            data: bytes.to_vec(),
+        }
+    }
```
storopoli, line 53: Should at least check if the `type` is valid.

storopoli (changes requested):

crates/predicate/src/key.rs
delbonis: It probably makes sense to have multiple different types for different representations.  One for what you describe when constructing things and another like this for when parsing things zero-copy and retaining the ability to get hashes of them easily:

```rs
pub struct PredicateKeyBuf<'b> {
    buf: &'b [u8]
}
```

crates/predicate/src/verifiers/mod.rs
delbonis, line 1: Split all these defs into other modules.

crates/predicate/src/verifiers/mod.rs
delbonis: It's not for testing *the library*, it's for testing things that use the predicate spec.

crates/predicate/src/verifiers/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::errors::PredicateError;
+use crate::verifier::PredicateVerifier;
+
+/// Enum representing all supported predicate types and their implementations.
+///
+/// This enum serves as a dispatch mechanism to route verification calls to the
+/// appropriate predicate-specific implementation based on the predicate type.
+/// Each variant contains the implementation for a specific predicate backend.
+#[derive(Debug)]
+pub(crate) enum PredicateImpl {
```
delbonis, line 25: Should this be `PredicateType`?  I can see why we might have multiple types that use the same sp1 verifier (like new master circuit vks) which would be presented as multiple variants with the same inner type.

Maybe `VerifierType`?

crates/predicate/src/verifiers/sp1_groth16.rs
```diff
@@ -0,0 +1,233 @@
...
+/// - `program_id`: Identifier of the SP1 program
+/// - `hash(public_values)`: SHA-256 or Blake3 hash of the claim
+#[derive(Debug, Default)]
+pub(crate) struct Sp1Groth16VerifierImpl;
+
+impl PredicateVerifier for Sp1Groth16VerifierImpl {
+    type Predicate = Groth16VerifyingKey;
+    type Witness = Groth16Proof;
+
+    fn parse_condition(&self, condition: &[u8]) -> Result<Self::Predicate> {
```
delbonis, line 43: Shouldn't this `Predicate` field be `Condition`?

crates/predicate/src/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+        reason: String,
+    },
+
+    // === Verification Errors ===
+    /// Predicate validation failed during verification setup.
+    #[error("predicate validation failed: {reason}")]
+    ValidationFailed {
+        /// The reason for validation failure.
+        reason: String,
+    },
```
delbonis, line 38: See if these reasons could be turned into a `&'static str` so that we don't have to alloc.

crates/predicate/src/errors.rs
```diff
@@ -0,0 +1,58 @@
...
+
+    /// Witness verification failed - witness does not satisfy the predicate.
+    #[error("witness verification failed")]
+    VerificationFailed,
+
+    // === Serialization Errors ===
+    /// Serialization or deserialization error.
+    #[error("serialization error: {0}")]
+    SerializationError(String),
```
delbonis, line 47: Maybe feature flag this one out so that it's not present unless signing is enabled?  Would be good to avoid this being needed in slim contexts.

crates/predicate/src/verifier.rs
delbonis: It's an implementation detail, predicates keys are opaque.

crates/predicate/src/constants.rs
```diff
@@ -0,0 +1,16 @@
...
+pub const NEVER_ACCEPT_PREDICATE_TYPE: PredicateType = 0;
+
+/// Always accepts any witness for any claim predicate type.
+pub const ALWAYS_ACCEPT_PREDICATE_TYPE: PredicateType = 1;
+
+/// Schnorr signature verification using BIP-340 predicate type.
+pub const BIP340_SCHNORR_PREDICATE_TYPE: PredicateType = 10;
+
+/// SP1 Groth16 verifier program verification predicate type.
+pub const SP1_GROTH16_PREDICATE_TYPE: PredicateType = 20;
```
delbonis, line 16: I feel like this could be a

```
#[repr(u8)]`
pub enum PredicateTypeId {
    ...
}
```

crates/predicate/src/key.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// * `claim` - The claim bytes to verify against
+    /// * `witness` - The witness bytes to verify
+    ///
+    /// # Returns
+    /// * `Ok(())` if verification succeeds
+    /// * `Err(PredicateError)` if verification fails or an error occurs
+    ///
+    /// [`verify_claim_witness`]: crate::verify_claim_witness
+    pub fn verify_claim_witness(&self, claim: &[u8], witness: &[u8]) -> Result<()> {
+        crate::verify_claim_witness(self, claim, witness)
```
delbonis, line 107: Can't we just call this directly?

crates/predicate/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+///
+/// # Arguments
+/// * `predicate` - The predicate key containing the type and condition data
+/// * `claim` - The claim bytes to verify against
+/// * `witness` - The witness bytes to verify
+///
+/// # Returns
+/// * `Ok(())` if verification succeeds
+/// * `Err(PredicateError)` if verification fails or an error occurs
+pub fn verify_claim_witness(predicate: &PredicateKey, claim: &[u8], witness: &[u8]) -> Result<()> {
```
delbonis, line 129: Should be another impl of this that operates all on byte slices so consumers aren't made to parse the predicate key and just returns true/false.

Also should have a function to validate a predicate key without actually parsing it, as referenced in the spec.

crates/predicate/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+k256 = { version = "0.13.4", features = ["schnorr"] }
+signature = "2.2"
+thiserror.workspace = true
+zkaleido-sp1-groth16-verifier = { git = "https://github.com/alpenlabs/zkaleido", branch = "feat/expose-types" }
+
+[dev-dependencies]
+rand.workspace = true
+sp1-verifier = "5.2.1"
+zkaleido = { git = "https://github.com/alpenlabs/zkaleido", branch = "feat/expose-types" }
+hex.workspace = true
```
delbonis, line 20: isn't the linter mad at you for these not being sorted?

delbonis (changes requested):
overall looks good, would like to see some allocs avoided

crates/predicate/Cargo.toml
prajwolrg: Updated the taplo config to fix this at [b50b67a](https://github.com/alpenlabs/strata-common/pull/15/commits/b50b67a1f76316193fa814af84fef98630dae19a).

crates/predicate/src/errors.rs
prajwolrg: This is actually not needed.

delbonis (approved):

crates/predicate/src/verifiers/sp1_groth16.rs
```diff
@@ -0,0 +1,210 @@
...
+            reason: format!("failed to compute Blake3 hash of claim: {e}"),
+        })?;
+
+        // Retry verification with Blake3 hash using zkaleido's verifier
+        verify_sp1_groth16_algebraic(program, proof, &fr_blake3).map_err(|e| {
+            PredicateError::VerificationFailed {
+                id: PredicateTypeId::Sp1Groth16,
+                reason: format!("SP1 Groth16 proof verification failed with both SHA-256 and Blake3 hashing: {e}"),
+            }
+        })
```
AaronFeickert, line 94: This feels sketchy to me. Is there a case where we'd reasonably want this behavior, and where the caller wouldn't know which hash to expect? Seems safer to make the two hash cases separate formats.

crates/predicate/src/verifiers/sp1_groth16.rs
```diff
@@ -0,0 +1,210 @@
...
+    use super::*;
+    use crate::test_utils::{
+        assert_predicate_parsing_failed, assert_verification_failed, assert_witness_parsing_failed,
+    };
+    use sp1_verifier::GROTH16_VK_BYTES;
+    use zkaleido::ProofReceiptWithMetadata;
+    use zkaleido_sp1_groth16_verifier::SP1Groth16Verifier;
+
+    fn load_predicate_claim_witness() -> (Vec<u8>, Vec<u8>, Vec<u8>) {
+        let program_id_hex = "00eb7fd5709e4b833db86054ba4acca001a3aa5f18b7e7d0d96d0f1d340b4e34";
```
AaronFeickert, line 109: How was this generated?

crates/predicate/src/verifiers/sp1_groth16.rs
```diff
@@ -0,0 +1,210 @@
...
+        // Test with larger witness
+        let mut larger_witness = witness.clone();
+        larger_witness.extend_from_slice(&[0u8; 10]);
+        let res = verifier.verify(&predicate, &claim, &larger_witness);
+        assert_witness_parsing_failed(res, PredicateTypeId::Sp1Groth16);
+
+        // Test with shorter witness
+        let shorter_witness = &witness[..witness.len() - 5];
+        let res = verifier.verify(&predicate, &claim, shorter_witness);
+        assert_witness_parsing_failed(res, PredicateTypeId::Sp1Groth16);
```
AaronFeickert, line 181: Why not test with single added and missing bytes as well?

crates/predicate/src/verifiers/sp1_groth16.rs
```diff
@@ -0,0 +1,210 @@
...
+        // Test with larger claim
+        let mut larger_claim = claim.clone();
+        larger_claim.extend_from_slice(&[0u8; 10]);
+        let res = verifier.verify(&predicate, &larger_claim, &witness);
+        assert_verification_failed(res, PredicateTypeId::Sp1Groth16);
+
+        // Test with shorter claim
+        let shorter_claim = &claim[..claim.len() - 2];
+        let res = verifier.verify(&predicate, shorter_claim, &witness);
+        assert_verification_failed(res, PredicateTypeId::Sp1Groth16);
```
AaronFeickert, line 203: Why not test with single added and missing bytes as well?

crates/predicate/src/key.rs
```diff
@@ -0,0 +1,298 @@
...
+            return Ok(Self {
+                id: PredicateTypeId::NeverAccept,
+                condition: &[],
+            });
+        }
+
+        let id = PredicateTypeId::try_from(bytes[0])?;
+        Ok(Self {
+            id,
+            condition: &bytes[1..],
```
AaronFeickert, line 120: Won't this panic on a single-byte slice?

crates/predicate/src/key.rs
```diff
@@ -0,0 +1,298 @@
...
+impl<'b> TryFrom<&'b [u8]> for PredicateKeyBuf<'b> {
+    type Error = PredicateError;
+
+    /// Creates a predicate key buffer from borrowed bytes with validation.
+    ///
+    /// The format is: `[id: u8][condition: bytes…]`. An empty byte array represents the
+    /// "never accept" predicate. The first byte (when present) is validated against the
+    /// [`PredicateTypeId`] registry and the remaining bytes are exposed as the predicate
+    /// condition payload.
+    fn try_from(bytes: &'b [u8]) -> Result<Self, Self::Error> {
```
AaronFeickert, line 109: Recommend fuzzing this parser.

crates/predicate/src/type_ids.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    fn try_from(value: u8) -> Result<Self, Self::Error> {
+        match value {
+            0 => Ok(PredicateTypeId::NeverAccept),
+            1 => Ok(PredicateTypeId::AlwaysAccept),
+            10 => Ok(PredicateTypeId::Bip340Schnorr),
+            20 => Ok(PredicateTypeId::Sp1Groth16),
+            invalid => Err(PredicateError::InvalidPredicateType(invalid)),
+        }
+    }
```
AaronFeickert, line 76: Recommend adding round-trip testing of these, since they're hardcoded separately.

AaronFeickert: Alternatively, could use a simple fuzzer for the parser to test for canonical representations.

AaronFeickert:
@storopoli almost done with a PR that addresses the recommendations, with the exception of the per-hash SP1 idea.

AaronFeickert:
See #24 for a PR that addresses most of the comments.

crates/predicate/src/verifiers/schnorr.rs
```diff
@@ -0,0 +1,215 @@
...
+    fn test_schnorr_signature_verification_pass() {
+        let verifier = SchnorrVerifier;
+
+        // Generate random message
+        let mut msg = [0u8; 32];
+        OsRng.fill_bytes(&mut msg);
+
+        // Create a modified message to test failure cases
+        let mut mod_msg = msg;
+        mod_msg.swap(1, 2);
```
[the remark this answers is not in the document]
AaronFeickert: Addressed in #24.

crates/predicate/src/verifiers/schnorr.rs
```diff
@@ -0,0 +1,215 @@
...
+        let verifier = SchnorrVerifier;
+
+        // Test with larger witness
+        let mut larger_witness = witness.clone();
+        larger_witness.extend_from_slice(&[0u8; 10]);
+        let res = verifier.verify(&predicate, &claim, &larger_witness);
+        assert_witness_parsing_failed(res, PredicateTypeId::Bip340Schnorr);
+
+        // Test with shorter witness
+        let shorter_witness = &witness[..witness.len() - 5];
```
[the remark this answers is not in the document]
AaronFeickert: Addressed in #24.

crates/predicate/src/verifiers/schnorr.rs
```diff
@@ -0,0 +1,215 @@
...
+        let verifier = SchnorrVerifier;
+
+        // Test with larger claim
+        let mut larger_claim = claim.clone();
+        larger_claim.extend_from_slice(&[0u8; 10]);
+        let res = verifier.verify(&predicate, &larger_claim, &witness);
+        assert_verification_failed(res, PredicateTypeId::Bip340Schnorr);
+
+        // Test with shorter claim
+        let shorter_claim = &claim[..claim.len() - 2];
```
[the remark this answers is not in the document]
AaronFeickert: Addressed in #24.

crates/predicate/src/type_ids.rs
AaronFeickert: Addressed in #24.

crates/predicate/src/verifiers/sp1_groth16.rs
AaronFeickert: Addressed in #24.

crates/predicate/src/verifiers/sp1_groth16.rs
AaronFeickert: I see that this basically matches the [behavior](https://github.com/alpenlabs/zkaleido/blob/3ff1a395adf49876df4164494ea7d7e890602d5e/adapters/sp1/groth16-verifier/src/verifier.rs#L92-L95) of `zkaleido`. But the same question applies there!

PR #17, opened 2025-09-24
Codec crate

delbonis:
## Description

This will be useful in some of the notpackage chain stuffs in the main repo.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

### DIFF
--- a/crates/codec/Cargo.toml
+++ b/crates/codec/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "strata-codec"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+thiserror.workspace = true
+
+[lints]
+workspace = true
--- a/crates/codec/README.md
+++ b/crates/codec/README.md
@@ -1,0 +1,14 @@
+# strata-codec
+
+This is a lightweight utility library for encoding binary messages.  It's
+oriented towards small messages that comfortably fit in memory and assumes that
+we're always reading out of a flat buferr (ie. that we're not reading from a
+`Read`).
+
+We currently use `borsh` to accomplish these tasks, but it's not well-suited for
+the task and we are incrementally trying to migrate to either SSZ for general
+types or, in this case, a more explicitly-defined encoding scheme.
+
+This is essentially a fork of the codec scheme I used in another project of
+mine, being maintained independently so we can iterate on it to aid in things
+like making it more zero-copy friendly.
--- a/crates/codec/src/lib.rs
+++ b/crates/codec/src/lib.rs
@@ -1,0 +1,15 @@
+//! Special purpose binary encoding format.
+
+mod errors;
+pub use errors::CodecError;
+
+mod types;
+pub use types::{BufDecoder, Codec, Decoder, Encoder};
+
+mod macros;
+
+mod util;
+pub use util::{decode_buf_exact, encode_to_vec};
+
+#[cfg(test)]
+mod tests;
--- a/crates/codec/src/types.rs
+++ b/crates/codec/src/types.rs
@@ -1,0 +1,124 @@
+use crate::CodecError;
+
+/// Generic codec trait for "plain old data" types that compactly go between bytes.
+pub trait Codec: Sized {
+    /// Decodes self from a decoder.
+    fn decode(dec: &mut impl Decoder) -> Result<Self, CodecError>;
+
+    /// Encodes self into an encoder.
+    fn encode(&self, enc: &mut impl Encoder) -> Result<(), CodecError>;
+}
+
+/// Generic decoder trait that reads inputs.
+pub trait Decoder {
+    /// Reads a variable-size buf.  This does NOT include length tagging.
+    fn read_buf(&mut self, into: &mut [u8]) -> Result<(), CodecError>;
+
+    /// Reads a fixed size buf.  This does NOT include length tagging.
+    fn read_arr<const N: usize>(&mut self) -> Result<[u8; N], CodecError>;
+}
+
+/// Generic encoder trait that writes outputs.
+pub trait Encoder {
+    /// Writes a buf.  This does NOT include length tagging.
+    fn write_buf(&mut self, buf: &[u8]) -> Result<(), CodecError>;
+}
+
+/// Impl for byte arrays.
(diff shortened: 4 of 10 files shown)


crates/codec/README.md
```diff
@@ -0,0 +1,14 @@
+# strata-codec
+
+This is a lightweight utility library for encoding binary messages.  It's
+oriented towards small messages that comfortably fit in memory and assumes that
+we're always reading out of a flat buferr (ie. that we're not reading from a
```
storopoli, line 5:
```suggestion
we're always reading out of a flat buffer (ie. that we're not reading from a
```

crates/codec/README.md
```diff
@@ -0,0 +1,14 @@
...
+`Read`).
+
+We currently use `borsh` to accomplish these tasks, but it's not well-suited for
+the task and we are incrementally trying to migrate to either SSZ for general
+types or, in this case, a more explicitly-defined encoding scheme.
+
+This is essentially a fork of the codec scheme I used in another project of
+mine, being maintained independently so we can iterate on it to aid in things
+like making it more zero-copy friendly.
```
storopoli, line 14: Who is "I", either sign it or say "I, Trey Delbonis, ..."

crates/codec/Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[package]
+name = "strata-codec"
+version = "0.1.0"
+edition = "2024"
```
storopoli, line 4:
```suggestion
edition = "2024"
readme = "README.md"
```

Since there's very relevant information in the `README.md`

crates/codec/src/lib.rs
```diff
@@ -0,0 +1,15 @@
+//! Special purpose binary encoding format.
```
storopoli, line 1: This could also go into the `Cargo.toml`'s [`description`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-description-field) field.

crates/codec/src/types.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl<B: AsRef<[u8]>> BufDecoder<B> {
+    /// Constructs a new instance.
+    pub fn new(buf: B) -> Self {
+        Self { buf, at: 0 }
+    }
+
+    /// Returns the length of the underlying buffer.
+    pub fn len(&self) -> usize {
```
storopoli, line 52: this can be `const` no?

crates/codec/src/types.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.buf.as_ref().len()
+    }
+
+    /// Returns if the underlying buffer is empty.
+    pub fn is_empty(&self) -> bool {
```
storopoli, line 57: This can also be `const`

crates/codec/src/types.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.buf.as_ref().is_empty()
+    }
+
+    /// Returns the total number of remaining bytes that can be read.
+    pub fn remaining(&self) -> usize {
```
storopoli, line 62: Also `const`?

crates/codec/src/types.rs
storopoli, line 1: Are you reimplementing [`Cursor`](https://doc.rust-lang.org/std/io/struct.Cursor.html) and/or [`Seek`](https://doc.rust-lang.org/std/io/trait.Seek.html)?

crates/codec/src/types.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+impl_int_codec!(u8 1);
+impl_int_codec!(i8 1);
+impl_int_codec!(u16 2);
+impl_int_codec!(i16 2);
+impl_int_codec!(u32 4);
+impl_int_codec!(i32 4);
+impl_int_codec!(u64 8);
+impl_int_codec!(i64 8);
```
storopoli, line 122: What about `usize`s?

crates/codec/src/types.rs
```diff
@@ -0,0 +1,122 @@
...
+                let arr: [u8; $bytes] = dec.read_arr()?;
+                Ok(<$ity>::from_be_bytes(arr))
+            }
+
+            fn encode(&self, enc: &mut impl Encoder) -> Result<(), CodecError> {
+                enc.write_buf(&self.to_be_bytes())
+            }
+        }
+    };
+}
```
storopoli, line 113: Shouldn't this be in `macros.rs`?

storopoli (changes requested):
Some comments.
Most important, I'm not gonna approve this PR until we have [`protests!`](https://docs.rs/proptest/latest/proptest/macro.proptest.html) in-place. We need proper testing here.

crates/codec/src/types.rs
delbonis: how many bytes should it be? :)

crates/codec/src/types.rs
delbonis: This isn't publicly exposed, it's just to make it simpler to define the trait impl for the int types, so I didn't want to complicate the module hierarchy.  I can add a note.

crates/codec/src/types.rs
delbonis: Cursor/Seek are both too powerful (allows operations knowing the buffer position that I didn't want to expose) and too limiting (it's designed to also support wrapping arbitrary `Read`s).

Maybe I could still define a `Decoder` on `Cursor<B> where B: AsRef<[u8]>` or something but I'm not sure what the value of that would be.  I'll mess with it a little bit.

storopoli: I just want to see if code can be simplified. Less to maintain is always better.

crates/codec/src/types.rs
delbonis: `.as_ref()` isn't const

delbonis:
@storopoli you're gonna love this property testing

storopoli (approved):
I think this is fine, amazing job!
One thing if you really want to, not needed, is to avoid creating a new `codec-tests` crate and just move the `proptest!` stuff to a feature-gated and `cfg(test)` module. That will make easier in the future.

PR #88, opened 2026-04-08
feat(sp1-groth16-verifier): support proofs prefixed with vk

prajwolrg:
## Description

SP1 prepends a 4-byte VK hash tag to Groth16 proofs by default. This tag allows early verification that the proof was generated with the expected verifying key.

In our case, the `ProgramId` is already embedded directly into the merged Groth16 VK, so this prefix serves no purpose for us. However, since SP1 includes it by default, callers would need to manually strip those 4 bytes before passing the proof to our verifier — which is cumbersome and error-prone.

This PR adds support for accepting proofs with the prefix intact (both compressed and uncompressed formats). The prefix bytes are simply stripped during parsing rather than validated, providing seamless compatibility without requiring callers to modify their proofs.

### Type of Change

- [X] New feature/Enhancement (non-breaking change which adds functionality or enhances an existing one)
- [ ] Documentation update
- [ ] Refactor
- [X] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

- [X] I have performed a self-review of my code.
- [X] I have commented my code where necessary.
- [X] I have updated the documentation if needed.
- [X] My changes do not introduce new warnings.
- [X] I have added tests that prove my changes are effective or that my feature works.
- [X] New and existing tests pass with my changes.

### DIFF
--- a/crates/predicate/Cargo.toml
+++ b/crates/predicate/Cargo.toml
@@ -36,7 +36,7 @@ sp1-verifier = "5.2.1"
 zkaleido = { git = "https://github.com/alpenlabs/zkaleido", tag = "v0.1.0-alpha-rc20" }
 
 [features]
-default = ["serde", "borsh"]
+default = ["serde", "borsh", "sp1-groth16"]
 serde = ["dep:serde", "dep:hex"]
 borsh = ["dep:borsh"]
 arbitrary = ["dep:arbitrary"]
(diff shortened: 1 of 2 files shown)


MdTeach (approved):

irnb (approved):

storopoli (commented):
ACK ff620d1f54848283e63fcfee32acd38e333ad650

PR #97, opened 2026-04-28
feat(logging): make noisy-dep filter directives consumer-supplied

storopoli:
## Description

Extracts `EnvFilter` construction into a dedicated `build_env_filter` helper and adds `LoggerConfig::extra_filter_directives` (plus a `with_extra_filter_directives` builder and a matching `LoggingInitConfig::extra_filter_directives` field) so binaries pass directives like `sp1_core_executor=warn` or `jsonrpsee_server::server=warn` themselves rather than having those hardcoded inside the crate.

`RUST_LOG` is still appended last and wins on conflicts.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

Per review feedback on the previous revision: noisy-dependency suppression directives now belong to the consumer, not this crate. Downstream binaries that previously relied on the implicit `sp1_core_executor=warn` / `jsonrpsee_server::server=warn` defaults must now opt in by setting `extra_filter_directives` on `LoggerConfig` or `LoggingInitConfig`.

`LoggingInitConfig` gained a new required field (`extra_filter_directives: &[&str]`), so call sites need a small update (pass `&[]` to keep the previous behavior of "no extras").

Needs a new tag after merged.

## Checklist

## Related Issues

STR-3236

### DIFF
--- a/crates/logging/src/manager.rs
+++ b/crates/logging/src/manager.rs
@@ -1,5 +1,6 @@
 //! Logging initialization and shutdown management.
 
+use std::env;
 use std::sync::OnceLock;
 
 use metrics_exporter_otel::OpenTelemetryRecorder;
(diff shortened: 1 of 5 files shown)


crates/logging/src/manager.rs
```diff
@@ -123,6 +120,20 @@ pub fn init(config: LoggerConfig) {
     );
 }
 
+fn build_env_filter(env_filter: Option<&str>) -> EnvFilter {
+    let default_directives = "sp1_core_executor=warn,jsonrpsee_server::server=warn";
```
prajwolrg: This feels something that should probably be passed by the consumer of the crates.

storopoli: Yes, makes sense instead of hardcoding.
Check f254003980ae01af144e93060df99905f853e89f

delbonis (approved):

bewakes (approved):

storopoli:
Rebased and will merge

PR #104, opened 2026-05-21
feat(metrics): add process metrics initializer

voidash:
## Summary
- add a new `strata-metrics` crate for process-level metrics recorder setup
- keep `strata-logging` scoped to logging/tracing subscriber setup and trace OTLP
- move the span busy/idle `MetricsLayer` from `strata-logging` to `strata-metrics`
- add optional metrics exporter modes: disabled, OTLP, Prometheus `/metrics`, or OTLP+Prometheus fanout
- add a generic extra tracing-layer hook to `strata-logging` so binaries can install `strata_metrics::MetricsLayer` without coupling the crates
- restore `strata-service` to its existing OpenTelemetry instrumentation path; no service histogram bucket semantics change in this PR

## Verification
- cargo test -p strata-logging -p strata-metrics -p strata-service
- cargo check -p strata-logging -p strata-metrics -p strata-service --locked
- vertex-core: cargo check -p strata -p alpen-client -p strata-signer-bin with local patches for strata-logging, strata-metrics, strata-service, and strata-tasks

### DIFF
--- a/crates/metrics/Cargo.toml
+++ b/crates/metrics/Cargo.toml
@@ -1,16 +1,19 @@
 [package]
 edition = "2024"
-name = "strata-logging"
+name = "strata-metrics"
 version = "0.1.0"
 
 [lints]
 workspace = true
 
 [dependencies]
+metrics.workspace = true
+metrics-exporter-otel.workspace = true
+metrics-exporter-prometheus.workspace = true
+metrics-util.workspace = true
 opentelemetry.workspace = true
 opentelemetry-otlp.workspace = true
 opentelemetry_sdk.workspace = true
+tokio.workspace = true
 tracing.workspace = true
-tracing-appender.workspace = true
-tracing-opentelemetry.workspace = true
 tracing-subscriber.workspace = true
--- a/crates/metrics/Cargo.toml
+++ b/crates/metrics/Cargo.toml
@@ -6,14 +6,26 @@ version = "0.1.0"
 [lints]
 workspace = true
 
+[features]
+default = ["otlp", "prometheus"]
+otlp = [
+  "dep:metrics-exporter-otel",
+  "dep:metrics-util",
+  "dep:opentelemetry",
+  "dep:opentelemetry-otlp",
+  "dep:opentelemetry_sdk",
+]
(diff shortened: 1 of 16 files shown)


crates/metrics/src/manager.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    recorder
+}
+
+fn build_prometheus_recorder(
+    listen_addr: SocketAddr,
+) -> metrics_exporter_prometheus::PrometheusRecorder {
+    tokio::runtime::Handle::try_current()
+        .expect("metrics init: Prometheus metrics exporter requires an active Tokio runtime");
```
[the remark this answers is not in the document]
delbonis: Could address this by making it take a runtime `Handle` as an arg and launching the exporter on that instead of creating an implicit dependency on one already being launched on the current thread.

crates/metrics/src/types.rs
```diff
@@ -0,0 +1,207 @@
...
+        /// Address used by the Prometheus HTTP listener.
+        listen_addr: SocketAddr,
+    },
+    /// Export metrics through OTLP and expose them through Prometheus.
+    OtlpAndPrometheus {
+        /// Address used by the Prometheus HTTP listener.
+        listen_addr: SocketAddr,
+    },
+}
+
```
delbonis: Maybe it'd be better to make this be a struct with fields for otlp and prometheus?  If we added another metrics config option then we'd have to add 4 more variants for each combination.

a la

```rs
pub struct MetricsConfig {
    otlp_endpoint: Option<String>,
    prometheus_listener_addr: Option<SocketAddr>,
}
```

crates/metrics/Cargo.toml
delbonis, line 1: This isn't really necessary for our use case, but it'd be "good practice" to make the different exporter backends (otlp, prometheus, etc.) feature flagged and disable the various code paths (ie. make the config trigger an error if set) when they're compiled out.  We can have them all enabled by default.

delbonis (changes requested):
Very good and clean, some comments on usage.

delbonis (approved):

PR #107, opened 2026-06-01
feat(merkle): add node-storing MMR behind `node_store` feature

prajwolrg:
## Description

Ports the reusable node store from the asm repo's `strata-mmr-store` into strata-merkle so it can be shared. Persisting every MMR node lets inclusion proofs be generated in O(log n) by walking the stored sibling path, with no leaf replay.

Reworked to be generic over the hasher rather than hardcoding Buf32/Sha256: the `MmrNodeStore` backend carries an associated `Hash`, and the derived `StoredMmr<MH>` API combines nodes via the supplied `MerkleHasher`, mirroring the existing `MmrState`/`Mmr` split. Proofs come back as the generic `MerkleProof` and verify against the same compact-peaks accumulators. The feature is opt-in and pulls in no new dependencies.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

Companion to https://github.com/alpenlabs/asm/pull/121.

**Where this fits, and where it doesn't.** This node store is deliberately minimal: one backend instance == one MMR, node hashes only (no preimages), append/overwrite with no deletes or compare-and-set. That's the right shape for a dedicated, append-only MMR that owns its own tree — e.g. the ASM manifest MMR, where the asm `strata-mmr-store` consumer maps one sled tree directly onto one `MmrNodeStore`.

It is **not** a drop-in for alpen, and intentionally so. Alpen keeps many MMRs (the ASM manifest MMR plus per-account snark-message-inbox MMRs) inside shared storage namespaced by `MmrId`, and explicitly does not want to open a separate tree per MMR. Its `MmrIndexDatabase` also stores leaf preimages, uses compare-and-set preconditions + atomic cross-MMR batches, and supports deletes/rollback — none of which the minimal `MmrNodeStore`/`StoredMmr` traits here model. So alpen should keep its DB-level trait; this crate is not meant to replace it.

## Checklist


crates/merkle/src/node_store/store.rs
```diff
@@ -0,0 +1,449 @@
...
+    /// `leaf_index` may be the current end (an append, which extends the leaf
+    /// count) or an existing index (an overwrite). The leaf, its recomputed
+    /// ancestors, and any leaf-count bump are written in a single
+    /// [`commit`](MmrNodeStore::commit).
+    ///
+    /// Errors with [`MmrError::NodeMissing`] if a required sibling is absent —
+    /// e.g. writing past the current end, which would leave a gap.
+    fn put_leaf(&self, leaf_index: u64, value: MH::Hash) -> Result<(), MmrError<Self::Error>> {
+        let old_count = <Self as StoredMmr<MH>>::leaf_count(self)?;
+        let new_count = old_count.max(leaf_index + 1);
```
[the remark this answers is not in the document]
evgenyzdanovich: hm, I'm not entirely sure, but it seems to me as a legit finding.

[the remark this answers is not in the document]
prajwolrg: @codex review

crates/merkle/src/node_store/store.rs
```diff
@@ -0,0 +1,487 @@
...
+        // Every real hash is at least 8 bytes; the count rides in the leading 8.
+        let mut bytes = [0u8; N];
+        bytes[..8].copy_from_slice(&value.to_le_bytes());
+        bytes
+    }
+
+    fn unpack_u64(&self) -> u64 {
+        let mut bytes = [0u8; 8];
+        bytes.copy_from_slice(&self[..8]);
+        u64::from_le_bytes(bytes)
```
delbonis: Shouldn't this be big-endian because of sorting reasons?

crates/merkle-store/src/index.rs
```diff
@@ -0,0 +1,231 @@
...
+        Self {
+            height: self.height + 1,
+            index: self.index >> 1,
+        }
+    }
+
+    /// Returns true if this node is the left child of its parent.
+    pub fn is_left_child(self) -> bool {
+        self.index.is_multiple_of(2)
+    }
```
delbonis, line 50: Should have a `is_right_child` for completeness.  Also I think the other impl had a `children` fn that returned a `Option<(NodePos, NodePos)>` with both children.  Just trying to be exhaustive with what downstream users might want to do.

crates/merkle/src/node_store/index.rs
```diff
@@ -0,0 +1,231 @@
...
+        Self(index)
+    }
+
+    /// Returns the zero-based leaf index.
+    pub fn index(self) -> u64 {
+        self.0
+    }
+
+    /// Returns this leaf's position in the node tree (height 0).
+    pub fn node_pos(self) -> NodePos {
```
delbonis: Should have `to_` or `as_` prefix.  Honestly maybe also add a `From`/`Into` impl to go from `LeafPos` to `NodePos`, as this is a good use case for that.

crates/merkle-store/src/index.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub fn node_pos(self) -> NodePos {
+        NodePos::new(0, self.0)
+    }
+}
+
+/// Size (in leaves) of the highest perfect mountain that fits in `leaves`.
+fn highest_mountain_size(leaves: u64) -> u64 {
+    debug_assert!(leaves > 0, "highest_mountain_size: leaves must be > 0");
+    1u64 << (63 - leaves.leading_zeros())
+}
```
delbonis, line 127: I think you might be able to do this with `next_power_of_2` and fewer bit tricks.

delbonis (changes requested):
A few minor ergonomic nits.  But also, *does* it make sense to include this in this crate?

crates/merkle/src/node_store/store.rs
```diff
@@ -0,0 +1,487 @@
...
+        // in `write_plan` don't always catch: an isolated height-0 peak (e.g.
+        // leaf 4 in a 5-leaf MMR) recomputes no ancestors, so the gap would
+        // commit silently. Reject the whole range explicitly.
+        if leaf_index > old_count {
+            return Err(MmrError::LeafGap {
+                index: leaf_index,
+                leaf_count: old_count,
+            });
+        }
+        let new_count = old_count.max(leaf_index + 1);
```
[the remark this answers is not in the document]
prajwolrg: Fixed in 1ef85cd: `put_leaf` now uses `leaf_index.checked_add(1).ok_or(MmrError::MaxCapacity)?`, so a full store (`u64::MAX` leaves) returns the new `MaxCapacity` error instead of overflowing into `write_plan`. Added `append_at_capacity_is_rejected` covering both `append_leaf` and a direct put at `u64::MAX`.

crates/merkle/src/node_store/store.rs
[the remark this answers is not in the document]
prajwolrg: Addressed in 7d86c05: `put_leaf` rejects `leaf_index > leaf_count` up front with `MmrError::LeafGap`, with a regression test for the isolated height-0 peak case (leaf 4 in a 5-leaf MMR) that the sibling reads alone would not catch.

crates/merkle/src/node_store/store.rs
prajwolrg: Switched to big-endian in 4fdd8a9. Strictly the count is stored as a *value* in the key->hash map and is never a sort key (only `NodePos::to_key` gets ordered), so endianness here does not change sorting — but I agree BE is the right call: it keeps this value encoded the same way as the key.

crates/merkle-store/src/index.rs
prajwolrg: Done in 502ea98: added `NodePos::is_right_child` and `children() -> Option<(NodePos, NodePos)>` (returns `None` for a height-0 leaf).

crates/merkle/src/node_store/index.rs
prajwolrg: Done in 502ea98: renamed `node_pos` -> `to_node_pos` and added `impl From<LeafPos> for NodePos`.

crates/merkle-store/src/index.rs
prajwolrg: Simplified in 502ea98 to `1 << leaves.ilog2()`. One note: `next_power_of_two` rounds *up* (smallest power >= n), but here we want the floor power (largest power <= leaves), so `ilog2` is the matching primitive — and it drops the `63 - leading_zeros()` trick.

MdTeach (approved):

delbonis (approved):

PR #116, opened 2026-06-05
refactor: rework SSZ impls for various wrapper-like types

delbonis:
## Description

This PR reworks SSZ encoding to use a more consistent "delegate" system.

I'm not really sure if this is the best approach to take here, but maybe we should expand this out some more.

**Some of these types seem like they're only used in the database, like `L1Payload`.**  So these maybe we should just remove the SSZ/borsh impls entirely and use Serde CBOR encodings for that, for forwards-compatibility.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

### DIFF
--- a/crates/btc-types/src/btc.rs
+++ b/crates/btc-types/src/btc.rs
@@ -22,10 +22,28 @@ use strata_codec::{Codec, CodecError, Decoder, Encoder};
 use strata_identifiers::{Buf32, SszDelegate, impl_ssz_transparent_wrapper, impl_ssz_via_delegate};
 
 use crate::ParseError;
-use crate::ssz_generated::ssz::btc::{BitcoinOutPointSsz, BitcoinScriptSsz, BitcoinTxOutSsz};
+use crate::ssz_generated::ssz::btc::{
+    BitcoinOutPointSsz, BitcoinScriptSsz, BitcoinTxOutSsz, MAX_SCRIPT_SIZE,
+};
 
 const HASH_SIZE: usize = 32;
 const BITCOIN_TXID_LEN: usize = 32;
+
+/// Validates that a Bitcoin script fits within the SSZ-encodable bound
+/// (`MAX_SCRIPT_SIZE`).
+///
+/// This is the single chokepoint enforced by every fallible constructor and
+/// deserialization path of [`BitcoinTxOut`] and [`BitcoinScriptBuf`], so that an
+/// instance can never hold a script that would overflow the SSZ `script_pubkey`
+/// list and panic during encoding/tree-hashing.
+fn check_script_size(len: usize) -> Result<(), ParseError> {
+    let max = MAX_SCRIPT_SIZE as usize;
+    if len > max {
+        Err(ParseError::ScriptTooLarge { size: len, max })
+    } else {
+        Ok(())
+    }
+}
 
 /// L1 output reference.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Serialize, Deserialize)]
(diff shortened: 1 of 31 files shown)


prajwolrg (approved):
The changes LGTM!

> Some of these types seem like they're only used in the database, like L1Payload. So these maybe we should just remove the SSZ/borsh impls entirely and use Serde CBOR encodings for that, for forwards-compatibility.

I think these types should then be part of this crate altogether. There's also things like `GenesisL1View`, which isn't relevant anymore. I am thinking of reassessing all the types in `strata-btc-types` and have a better framework about what should be included here.

crates/btc-types/ssz/btc.ssz
```diff
@@ -0,0 +1,74 @@
+# SSZ schema definitions for strata-btc-types wrapper types.
+#
+# These containers are the canonical SSZ "delegate" representations for the
+# Bitcoin wrapper types in this crate. The Rust wrappers implement `SszDelegate`
+# against the generated types and obtain their `Encode`/`Decode` impls via
+# `impl_ssz_via_delegate!`, so the wire layout is correct by construction (the
+# SSZ fixed-part/variable-part split is handled entirely by the generated code).
+
+### Maximum size of a Bitcoin scriptPubKey, in bytes (Bitcoin Core `MAX_SCRIPT_SIZE`).
+MAX_SCRIPT_SIZE = 10000
```
storopoli, line 10: Do we accept `_` as thousands sep in `ssz-gen`?

```suggestion
MAX_SCRIPT_SIZE = 10_000
```

crates/btc-types/src/btc.rs
```diff
@@ -433,34 +392,35 @@ impl<'a> Arbitrary<'a> for BitcoinTxid {
...
+
+    fn into_delegate(self) -> Self::Delegate {
+        BitcoinTxOutSsz {
+            value: self.0.value.to_sat(),
+            script_pubkey: self
+                .0
+                .script_pubkey
+                .to_bytes()
+                .try_into()
+                .expect("scriptPubKey exceeds MAX_SCRIPT_SIZE"),
```
storopoli, line 449: These delegate conversions are still fallible for values the public API can construct. `BitcoinTxOut` can be built from any `TxOut` via `From<TxOut>` and Borsh decode, and `BitcoinScriptBuf` can be built from any `ScriptBuf`; if the script exceeds `MAX_SCRIPT_SIZE`, `as_ssz_bytes()`/tree hashing panics here. This should mirror `ThresholdConfig`: validate the bound at construction/deserialization and avoid infallible `From` for unchecked values, or choose a delegate representation that covers every value the wrapper can hold.

crates/crypto/src/threshold_signature/indexed/config.rs
```diff
@@ -337,47 +320,44 @@ impl<'a> Arbitrary<'a> for ThresholdConfigUpdate {
...
+impl SszDelegate for ThresholdConfigUpdate {
+    type Delegate = ThresholdConfigUpdateSsz;
+
+    fn into_delegate(self) -> Self::Delegate {
+        let to_bytes_list = |keys: Vec<CompressedPublicKey>| {
+            keys.into_iter()
+                .map(|key| FixedBytes(key.serialize()))
+                .collect::<Vec<_>>()
+                .try_into()
+                .expect("more than MAX_SIGNERS members")
```
storopoli, line 357: `ThresholdConfigUpdate::new` still accepts arbitrary add/remove lists, but the delegate is bounded to `MAX_SIGNERS`. A standalone update with more than 256 add or remove members can be constructed and then panic under `as_ssz_bytes()`/tree hashing at this `expect`. `ThresholdConfig` already makes this non-fallible by validating at construction/update time; the update type should do the same, or avoid exposing SSZ for unchecked updates.

storopoli (changes requested):

crates/btc-types/ssz/btc.ssz
delbonis: I don't think so, I can add that in a future revision.

crates/btc-types/src/btc.rs
delbonis: Added more bounds checks.

crates/crypto/src/threshold_signature/indexed/config.rs
delbonis: Added more bounds checks.

storopoli (approved):
ACK 68478a1dac3bf642684d3fafd4d09777d2c46f35

CONDITIONAL on the removal of 89389a0d60092f70e50145cc90304fff6f16764a

PR #117, opened 2026-06-07
feat(merkle-store): pruning for the node store

prajwolrg:
## Description

The node-storing MMR (`node_store` feature, added in #107) only ever grew: every node was retained forever, so a long-lived store had no way to bound its size or roll back. This PR adds pruning, keyed by leaf position.

- **`prune_before`** forgets old history: it keeps only the peaks of the leading prefix (the minimal set still needed to prove later leaves and to keep appending) and drops their descendants. The leaf count is left intact, and a monotonic `pruned_before` watermark is recorded so that proof generation below it returns the new `MmrError::Pruned` rather than `NodeMissing` (which signals a corrupt store).
- **`prune_after`** truncates the MMR to a prefix: it deletes a leaf and everything after it, updates the stored leaf count, and clamps the watermark down to the surviving leaf count. An earlier `prune_before` can have discarded the prefix peaks a truncation would need to stay appendable, so a cut below the watermark is probed and rejected with `MmrError::Pruned` rather than committing an unappendable store.
- **`pop_leaf`** removes just the last leaf and returns its value, the inverse of `append_leaf`. It is a thin convenience over `prune_after` at `leaf_count - 1`, so stack-like consumers can undo the most recent append without computing the index themselves.

Deletion is folded into `MmrNodeStore`: a backend now implements a third method, `delete_node`, and `commit` is widened to take both writes and deletes. A prune (node deletes + watermark/count update) then goes through the same single mutation path as a leaf write, so a transactional backend can make the whole batch atomic in one place.

Durability and atomicity are delegated to the backend's `commit`. Independently of that, the prune operations are idempotent: the watermark is monotonic, the delete sets recompute identically, and the count/watermark settle to the same values, so a consumer that re-runs a prune converges to the same state.

This is **not** bounded-memory pruning: each `prune_*` materializes its full delete set into a `Vec` before handing it to `commit`, so a single prune over a long-lived MMR allocates memory proportional to the number of nodes removed. Because the operations are idempotent, a consumer that needs to bound peak memory can prune in steps (advancing `before`, or truncating in chunks) and converge to the same state; bounding it that way is left to the consumer.

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

Intended use: `prune_before` reclaims storage by forgetting old history, `prune_after` removes entries on a reorg by truncating the MMR back to a known-good prefix, and `pop_leaf` undoes a single append.

Note that `MmrNodeStore` is now a three-method contract (`get_node` / `put_node` / `delete_node`) and `commit` takes `(writes, deletes)`, so any existing backend implementation needs updating. `MemMmr` is the only implementor in this repo.

Builds on #107. Suggested reading order:

1. `error.rs` — the new `MmrError::Pruned` variant.
2. `store.rs` — the widened `MmrNodeStore` (`delete_node` plus `commit(writes, deletes)`) and the `prune_before` / `prune_after` / `pop_leaf` methods.
3. `algorithm.rs` — `iter_prune_before_positions` / `iter_prune_after_positions`, which compute what to keep vs. drop.

## Checklist

## Related Issues

Builds on #107.

### DIFF
--- a/crates/merkle-store/src/algorithm.rs
+++ b/crates/merkle-store/src/algorithm.rs
@@ -87,44 +87,42 @@ pub fn assemble_proof<H: MerkleHash>(leaf_index: u64, cohashes: Vec<H>) -> Merkl
 /// Positions to delete to prune everything strictly before leaf `before`,
 /// keeping only the peaks of the first `before` leaves.
 ///
-/// Returns the proper descendants of each peak in [`peak_positions`]`(before)` —
+/// Yields the proper descendants of each peak in [`peak_positions`]`(before)` —
 /// exactly the nodes whose leaf-coverage lies entirely in `[0, before)`, minus
 /// the peaks themselves, which are retained so later leaves stay provable and
 /// appends keep working. Empty when `before == 0`.
-// TODO: materializes the full position list; for very large prunes this could
-// be an iterator so the backend can delete in bounded-memory chunks.
-pub fn prune_before_positions(before: u64) -> Vec<NodePos> {
-    let mut positions = Vec::new();
-    for peak in peak_positions(before) {
+///
+/// Produced lazily — only the peak list (`O(log before)`) is held, with the
+/// descendants yielded on demand. The prune op collects the full set before
+/// handing it to the node store to delete.
+pub fn prune_before_positions(before: u64) -> impl Iterator<Item = NodePos> {
+    peak_positions(before).into_iter().flat_map(|peak| {
         let peak_height = peak.height();
         let peak_index = peak.index();
-        for height in 0..peak_height {
+        (0..peak_height).flat_map(move |height| {
             // Each descendant level widens the peak's index by 2 per step down.
             let span = 1u64 << (peak_height - height);
-            for index in (peak_index * span)..((peak_index + 1) * span) {
-                positions.push(NodePos::new(height, index));
-            }
-        }
-    }
-    positions
+            (peak_index * span..(peak_index + 1) * span)
+                .map(move |index| NodePos::new(height, index))
+        })
+    })
 }
 
 /// Positions to delete to truncate an MMR from `leaf_count` leaves down to
 /// `keep` leaves.
 ///
-/// Returns every node present at `leaf_count` but not at `keep`: at each height
+/// Yields every node present at `leaf_count` but not at `keep`: at each height
 /// `h`, the indices in `[keep >> h, leaf_count >> h)` (a node `(h, i)` exists at
 /// a given count iff `i < count >> h`). Empty when `keep >= leaf_count`.
-// TODO: materializes the full position list; for very large prunes this could
-// be an iterator so the backend can delete in bounded-memory chunks.
-pub fn prune_after_positions(keep: u64, leaf_count: u64) -> Vec<NodePos> {
-    let mut positions = Vec::new();
-    let mut height = 0u8;
-    while (leaf_count >> height) > 0 {
-        for index in (keep >> height)..(leaf_count >> height) {
-            positions.push(NodePos::new(height, index));
-        }
-        height += 1;
-    }
-    positions
+///
+/// Produced lazily; the prune op collects the full set before handing it to the
+/// node store to delete. A real MMR tops out at height 63 (a height-64 node
+/// would need `2^64` leaves), so the height walk is capped there, which also
+/// avoids a `>> 64` shift overflow.
+pub fn prune_after_positions(keep: u64, leaf_count: u64) -> impl Iterator<Item = NodePos> {
+    (0u8..64)
+        .take_while(move |&height| (leaf_count >> height) > 0)
+        .flat_map(move |height| {
+            (keep >> height..leaf_count >> height).map(move |index| NodePos::new(height, index))
+        })
 }
--- a/crates/merkle-store/src/algorithm.rs
+++ b/crates/merkle-store/src/algorithm.rs
@@ -11,7 +11,7 @@
 use strata_merkle::{MerkleHash, MerkleHasher, MerkleProof};
 
 use super::error::MmrError;
-use super::index::{LeafPos, NodePos, peak_for_leaf};
+use super::index::{LeafPos, NodePos, peak_for_leaf, peak_positions};
 
 /// Computes the nodes to write when setting `leaf` at `leaf_index` in an MMR
 /// that has (or will have) `leaf_count` leaves.
@@ -82,4 +82,49 @@ pub fn proof_positions(leaf_index: u64, leaf_count: u64) -> Vec<NodePos> {
 /// order (the order returned by [`proof_positions`]).
 pub fn assemble_proof<H: MerkleHash>(leaf_index: u64, cohashes: Vec<H>) -> MerkleProof<H> {
     MerkleProof::from_cohashes(cohashes, leaf_index)
+}
+
+/// Positions to delete to prune everything strictly before leaf `before`,
+/// keeping only the peaks of the first `before` leaves.
+///
+/// Returns the proper descendants of each peak in [`peak_positions`]`(before)` —
+/// exactly the nodes whose leaf-coverage lies entirely in `[0, before)`, minus
+/// the peaks themselves, which are retained so later leaves stay provable and
+/// appends keep working. Empty when `before == 0`.
+// TODO: materializes the full position list; for very large prunes this could
+// be an iterator so the backend can delete in bounded-memory chunks.
+pub fn prune_before_positions(before: u64) -> Vec<NodePos> {
+    let mut positions = Vec::new();
+    for peak in peak_positions(before) {
+        let peak_height = peak.height();
+        let peak_index = peak.index();
+        for height in 0..peak_height {
+            // Each descendant level widens the peak's index by 2 per step down.
+            let span = 1u64 << (peak_height - height);
+            for index in (peak_index * span)..((peak_index + 1) * span) {
+                positions.push(NodePos::new(height, index));
+            }
+        }
+    }
+    positions
+}
+
+/// Positions to delete to truncate an MMR from `leaf_count` leaves down to
+/// `keep` leaves.
+///
+/// Returns every node present at `leaf_count` but not at `keep`: at each height
+/// `h`, the indices in `[keep >> h, leaf_count >> h)` (a node `(h, i)` exists at
+/// a given count iff `i < count >> h`). Empty when `keep >= leaf_count`.
+// TODO: materializes the full position list; for very large prunes this could
+// be an iterator so the backend can delete in bounded-memory chunks.
+pub fn prune_after_positions(keep: u64, leaf_count: u64) -> Vec<NodePos> {
+    let mut positions = Vec::new();
+    let mut height = 0u8;
+    while (leaf_count >> height) > 0 {
+        for index in (keep >> height)..(leaf_count >> height) {
+            positions.push(NodePos::new(height, index));
+        }
+        height += 1;
+    }
+    positions
 }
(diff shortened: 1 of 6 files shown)


crates/merkle-store/src/store.rs
```diff
@@ -87,6 +101,54 @@ pub trait MmrNodeStore {
...
+        Ok(())
+    }
+
+    /// Deletes a (possibly very large, lazily produced) stream of positions in
+    /// fixed-size batches, each flushed through [`delete_nodes`](Self::delete_nodes).
+    ///
+    /// This is how the pruning operations on [`StoredMmr`] discard nodes: it
+    /// bounds peak memory to one fixed-size batch regardless of how many nodes a
+    /// prune removes. The default need not be overridden.
+    fn delete_chunked(&self, positions: impl Iterator<Item = NodePos>) -> Result<(), Self::Error> {
```
delbonis: Unfortunately I think this makes this non-dyn-safe so it can't be used in the storage indirection.

Since it's just defined in terms of operations that already exist here, we could add it to the MMR index manager, right?

delbonis (changes requested):
There's quite a lot of stuff here being implemented *as* trait default impls, these should probably be reorganized so the trait is *just* about the storage ops and the MMR ops are implemented *around* it.

crates/merkle-store/src/store.rs
prajwolrg: Good catch — you're right, the `impl Iterator` param is what made it non-object-safe. Rather than relocate the batching, I dropped `delete_chunked` from the trait entirely in 394d23d: the prune ops now hand the whole position set to `delete_nodes(&[NodePos])` in one call, and whether/how to chunk is left to the backend (documented on `delete_nodes`) — that keeps the chunk-size policy out of the library and with the implementor that actually knows its storage limits.

With that gone, `PrunableNodeStore` and `MmrNodeStore` are object-safe again — confirmed `&dyn PrunableNodeStore<Hash = …, Error = …>` compiles.

(Where the MMR ops themselves should live — the "MMR index manager" — is the bigger reorg; replied on that in the main thread.)

crates/merkle-store/src/store.rs
```diff
@@ -87,6 +94,38 @@ pub trait MmrNodeStore {
...
+    /// a count that scales with how far back the cut reaches (a deep
+    /// [`prune_after`](StoredMmr::prune_after), for instance, can delete most of
+    /// the tree). A backend that cannot issue an arbitrarily large delete in one
+    /// go should chunk internally (e.g. `positions.chunks(N)`) to its own limit.
+    fn delete_nodes(&self, positions: &[NodePos]) -> Result<(), Self::Error> {
+        for pos in positions {
+            self.delete_node(*pos)?;
+        }
+        Ok(())
+    }
```
delbonis: This should just provide the batch delete operation as native one, that's the more likely scenario that we should be optimizing for.

I would also merge this with the other node store trait since it's also normal to have del operations where you have put operations.  Logical correctness can be ensured by callers, since we're only going to be exposing high-level MMR operations to consumers anyways.  This isn't a scenario where we get a lot out of meticulously enumerating the trait hierarchy.

crates/merkle-store/src/algorithm.rs
```diff
@@ -83,3 +83,46 @@ pub fn proof_positions(leaf_index: u64, leaf_count: u64) -> Vec<NodePos> {
...
+/// Yields the proper descendants of each peak in [`peak_positions`]`(before)` —
+/// exactly the nodes whose leaf-coverage lies entirely in `[0, before)`, minus
+/// the peaks themselves, which are retained so later leaves stay provable and
+/// appends keep working. Empty when `before == 0`.
+///
+/// Produced lazily — only the peak list (`O(log before)`) is held, with the
+/// descendants yielded on demand. The prune op collects the full set before
+/// handing it to the node store to delete.
+pub fn prune_before_positions(before: u64) -> impl Iterator<Item = NodePos> {
+    peak_positions(before).into_iter().flat_map(|peak| {
```
delbonis: Maybe we could rewite `peak_positions` to return an `impl Iterator` so we don't need this `into_iter` call?

Also the naming is a bit confusing, this fn suggests it's *doing* the prune.  They should both (and `prune_after_positions`) have `get_` or `iter_` prefixes.

delbonis (changes requested):
Couple minor comments.

crates/merkle-store/src/algorithm.rs
prajwolrg: Done in 34d2bc6. `peak_positions` now returns `impl Iterator<Item = NodePos>`, so `iter_prune_before_positions` drops the `.into_iter()`. And both helpers are renamed to `iter_prune_before_positions` / `iter_prune_after_positions` so the names read as position iterators rather than actions that perform a prune.

crates/merkle-store/src/store.rs
prajwolrg: Merged in 20fd938: dropped `PrunableNodeStore` and moved deletion onto `MmrNodeStore`, so the prune/pop ops no longer carry a separate trait bound.

One deviation from "make the batch the native op": I kept `delete_node` as the single-node primitive with `delete_nodes` as an overridable batch default, to mirror the other two pairs — `get_node`/`get_nodes` and `put_node`/`put_nodes` (I also renamed `commit` -> `put_nodes` for that symmetry). A backend with a native multi-delete still overrides `delete_nodes` for a single round-trip, so the batch path is there for the scenario you flagged; it just is not the sole required method. Happy to flip it to batch-only if you would rather optimize purely for that.

crates/merkle-store/src/store.rs
```diff
@@ -216,6 +307,102 @@ where
...
+        self.delete_chunked(prune_before_positions(before_index))
+            .map_err(MmrError::Backend)?;
+        // Raise the watermark last (monotonically), once the nodes are gone, so
+        // a leaf is never reported `Pruned` while its path is still present.
+        let watermark = <Self as StoredMmr<MH>>::pruned_before(self)?;
+        if before_index > watermark {
+            self.commit(&[(
+                NodePos::meta(PRUNED_BEFORE_TAG),
+                MH::Hash::pack_u64(before_index),
+            )])
```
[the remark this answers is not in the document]
prajwolrg: Good catch — fixed in 86546bb. `put_leaf` now reads the watermark and rejects an overwrite with `leaf_index < pruned_before` as `MmrError::Pruned`, before it can recompute a retained peak. Appends (`leaf_index == old_count`) are always at or above the watermark, so they are unaffected. Added a regression test (`put_leaf_below_watermark_is_rejected`) covering the 6-leaf / `prune_before(5)` / `put_leaf(4)` case from your example.

crates/merkle-store/src/store.rs
```diff
@@ -216,6 +289,115 @@ where
...
+        }
+        let positions: Vec<NodePos> = iter_prune_after_positions(keep, leaf_count).collect();
+        self.delete_nodes(&positions).map_err(MmrError::Backend)?;
+        // Lower the leaf count last so an interrupted prune is resumable. Clamp
+        // the watermark to the new count in the same write, since no leaf at or
+        // above `keep` survives to be reported `Pruned`.
+        let mut writes = vec![(NodePos::meta(NEXT_INDEX_TAG), MH::Hash::pack_u64(keep))];
+        let watermark = <Self as StoredMmr<MH>>::pruned_before(self)?;
+        if watermark > keep {
+            writes.push((NodePos::meta(PRUNED_BEFORE_TAG), MH::Hash::pack_u64(keep)));
```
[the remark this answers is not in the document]
prajwolrg: Fixed in ee385f450b50ca4498c584b48e0e67d290063821.

`prune_after` now probes the peaks of the resulting `keep`-leaf MMR before committing and rejects with `MmrError::Pruned` if any was already discarded, so it can no longer lower the count / clamp the watermark onto a prefix it can't append to. Your repro (4 leaves → `prune_before(4)` → `prune_after(3)`) now returns `Pruned { index: 3, pruned_before: 4 }` and leaves the store untouched (regression test `prune_after_into_pruned_prefix_is_rejected`).

One subtlety worth flagging: the gate is "are the prefix peaks still present", not simply `keep < watermark`. A cut below the watermark whose peaks survived is still valid and still allowed — e.g. `prune_before(5)` then `prune_after(4)` keeps peak `(2,0)`, which is exactly what a 4-leaf MMR needs, so that one proceeds and clamps the watermark to 4. The probe only runs when `keep < watermark`; at or above it the peaks are always retained.

`pop_leaf` inherits this through `prune_after`, so popping the last leaf out of a compacted-away prefix is now refused rather than silently corrupting the count (previously it dropped the count and left an unappendable store). Whether it errors depends on the cut's shape — a shorter prefix whose peaks survive still pops cleanly.

crates/merkle-store/src/store.rs
```diff
@@ -216,6 +289,115 @@ where
...
+    fn prune_before(&self, before: LeafPos) -> Result<(), MmrError<Self::Error>> {
+        let leaf_count = <Self as StoredMmr<MH>>::leaf_count(self)?;
+        let before_index = before.index();
+        if before_index > leaf_count {
+            return Err(MmrError::LeafOutOfRange {
+                index: before_index,
+                leaf_count,
+            });
+        }
+        let positions: Vec<NodePos> = iter_prune_before_positions(before_index).collect();
```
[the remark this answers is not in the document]
prajwolrg: Documented rather than chunked, in 5d0f47e88d933d8412e303c2383f423ad8657a2a.

The reason it isn't chunked in the library: it pulls against the atomic-commit design that landed in 1202836cdae18d165ed1f665bfdb561b0235b585. The prune ops now hand their whole delete set (plus the watermark/count write) to a single `commit(writes, deletes)` so a transactional backend can apply the prune as one unit. Flushing the iterator in fixed-size chunks would mean multiple commits per prune, giving up that atomicity, and the library has no way to know a given backend's batch/transaction limits to pick a chunk size.

So the doc on `prune_before`/`prune_after` now states that the delete set is materialized in full and points consumers at stepwise pruning to bound peak memory: the watermark is monotonic and the delete set idempotent, so repeated calls advancing `before` (or truncating in steps) converge to the same state while keeping each `commit` atomic. The consumer — which does know its backend's limits — decides whether and how to bound it.

Happy to revisit if you'd rather the trait expose a streaming/iterator delete path instead; that'd be a deliberate trade of per-prune atomicity for bounded memory.

crates/merkle-store/src/store.rs
```diff
@@ -193,6 +266,17 @@ where
             });
         }
 
+        // A leaf below the watermark had its path discarded by `prune_before`;
+        // report that intent rather than letting the missing-node read below
+        // surface as `NodeMissing`, which a caller reads as corruption.
+        let pruned_before = <Self as StoredMmr<MH>>::pruned_before(self)?;
```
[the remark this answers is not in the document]
prajwolrg: Fixed in cf3bf3b.

`generate_proof_at_size` now reads the live `leaf_count()` and rejects with `MmrError::LeafOutOfRange { index: at_leaf_count - 1, leaf_count }` when `at_leaf_count` exceeds it, before assembling anything. Your repro (5 leaves → `prune_after(4)` → `generate_proof_at_size(4, 5)`) now errors instead of returning the empty-path proof.

The subtlety worth recording: the hazard wasn't a malformed proof but a *valid* one. Leaf 4 is a lone height-0 peak at size 5, so its proof path is empty and slips past the `NodeMissing` check — the assembled proof still verifies against the (now-abandoned) size-5 root. So the store would vouch for a leaf it had deliberately rolled back via `prune_after`, with nothing to signal the state is gone. The guard makes it vend proofs only against sizes it still holds. Regression test `proof_past_truncated_size_is_rejected` covers both the lone-peak leaf (4) and a still-retained leaf (2) against the gone size, and confirms proofs against the retained size 4 still verify.

delbonis (approved):
Couple minor comments.

PR #118, opened 2026-06-09
Add `#[gen_proxy]` attr macro

delbonis:
## Description

This PR creates a `strata-db-macros` crate which exposes the `gen_proxy` attribute macro.  It generates a "proxy" type with associated code in a similar style to the objects created by the `inst_ops!` family of declaritive macros.  This flexibility allows for a more streamlined design that removes the "context" indirection and uses a type based on `JoinHandle` instead of a `tokio::sync::oneshot` channel to return values from blocking tasks.

In a future PR I will rework the storage and db crates in the main Strata repo to use this macro (defining them in the db-types crate and then reexporting from the storage crate with the `Ops` suffix.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

### DIFF
--- a/crates/db-macros/src/expand.rs
+++ b/crates/db-macros/src/expand.rs
@@ -189,7 +189,7 @@ fn is_self_sized_predicate(pred: &syn::WherePredicate) -> bool {
         tb.path
             .segments
             .last()
-            .map_or(false, |s| s.ident == "Sized")
+            .is_some_and(|s| s.ident == "Sized")
     })
 }
 
@@ -210,11 +210,10 @@ fn should_proxy_method(method: &TraitItemFn) -> bool {
     if sig.asyncness.is_some() || !sig.generics.params.is_empty() {
         return false;
     }
-    if let Some(wc) = &sig.generics.where_clause {
-        if wc.predicates.iter().any(is_self_sized_predicate) {
+    if let Some(wc) = &sig.generics.where_clause
+        && wc.predicates.iter().any(is_self_sized_predicate) {
             return false;
         }
-    }
     match sig.inputs.iter().next() {
         Some(FnArg::Receiver(recv)) if recv.reference.is_some() && recv.mutability.is_none() => {}
         _ => return false,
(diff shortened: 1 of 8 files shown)


crates/db-macros/src/expand.rs
```diff
@@ -0,0 +1,294 @@
...
+        let ident = &pat_ident.ident;
+        arg_names.push(ident.clone());
+        arg_decls.push(quote! { #ident: #ty });
+    }
+
+    // The return type must be a `Result`-shaped path; extract its success type.
+    let ReturnType::Type(_, ret_ty) = &sig.output else {
+        return None;
+    };
+    let success_ty = first_generic_type(ret_ty)?;
```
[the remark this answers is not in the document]
delbonis: This is fine, we want to assume all the return types are fallible so that we have something to turn `JoinError`s into.

crates/db-macros/src/expand.rs
```diff
@@ -0,0 +1,294 @@
...
+) -> Option<TokenStream> {
+    // An explicit `#[gen_proxy(skip)]` opts the method out of proxying entirely.
+    if method.attrs.iter().any(is_skip_attr) {
+        return None;
+    }
+
+    let sig = &method.sig;
+
+    // Only plain, non-generic, synchronous methods are proxied.
+    if sig.asyncness.is_some() || !sig.generics.params.is_empty() {
```
storopoli: This eligibility check still proxies `&self` methods that are intentionally constrained with `where Self: Sized`. Those methods are allowed on an object-safe trait, but they cannot be called through the generated `Arc<dyn Trait>`; the generated `_blocking`/`_chan` bodies later dispatch through `self.inner` / `inner.as_ref()`, so a trait with a helper like `fn helper(&self) -> DbResult<T> where Self: Sized` will make the whole macro expansion fail to compile. Please either skip methods whose where-clause requires `Self: Sized`, or emit a clear macro error for that case.

storopoli (changes requested):

crates/db-macros/src/expand.rs
delbonis: Ok I added more constraints and improved the logic around here.

PR #122, opened 2026-06-18
refactor(btc-types): wrap rust-bitcoin Amount in BitcoinAmount

prajwolrg:
## Description

`BitcoinAmount` previously re-implemented `bitcoin::Amount` as a raw `u64` newtype, carrying a hand-rolled surface of arithmetic helpers and constants. Its only real reason to exist is to attach the `Borsh`/`Arbitrary`/SSZ/`Codec` impls that the upstream `Amount` lacks. This wraps `bitcoin::Amount` directly so `Display`/`Debug` and all value semantics come from rust-bitcoin, leaving the wrapper as just those trait impls plus conversions and `Deref` to `Amount`.

SSZ/Borsh/`Codec` still encode the amount as a bare `u64` count of satoshis, so the on-wire/at-rest binary format is unchanged. `Display` now renders in BTC (via `Amount`'s `Display`) and `Debug` as `<n> SAT`, where the old `Display` printed the raw sat integer.

This also enforces a `MAX_MONEY` (21M BTC) invariant on untrusted input, via a single `check_amount` chokepoint mirroring the existing `check_script_size` pattern:

- `TryFrom<u64>` validates and rejects values above `Amount::MAX_MONEY`.
- Every deserialization path (serde, Borsh, SSZ, `Codec`) validates the decoded satoshi count and rejects over-max values.
- `Arbitrary` only generates in-range values.
- Wrapping an already-typed `Amount` stays an infallible `From`, trusting the caller (so the bound is enforced on parsed input, not as an absolute type-level guarantee).

### Type of Change

- [x] Refactor
- [x] New or updated tests

## Notes to Reviewers

- **Decode behavior change:** deserializing now rejects satoshi counts above 21M BTC where any `u64` was previously accepted. The binary encoding itself is unchanged (still a `u64`), so in-range values round-trip identically.
- **API surface removed:** the inherent constants (`ZERO`/`MIN`/`MAX`/`MAX_MONEY`/`SIZE`/`SATS_FACTOR`) and helpers (`to_sat`, `from_sat`, `from_int_btc`, `checked_*`, `saturating_*`, `is_zero`, `zero`, `sum`) are gone. `BitcoinAmount` is used only within this crate today, and value methods (`to_sat`, `checked_*`, …) now come from `Amount` through `Deref`. Arithmetic is done in `Amount` space; `saturating_*` has no `Amount` equivalent and was dropped.
- `Display` cannot be derived (no `derive_more`/`strum` dependency), so it stays a one-line manual forward to `Amount`.

## Checklist

## Related Issues

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/btc-types/src/btc.rs
+++ b/crates/btc-types/src/btc.rs
@@ -142,9 +142,8 @@ impl<'a> Arbitrary<'a> for BitcoinOutPoint {
 /// deserialization path of [`BitcoinAmount`], so that an instance can never hold
 /// a value beyond the money supply.
 fn check_amount(sats: u64) -> Result<(), ParseError> {
-    let max = Amount::MAX_MONEY.to_sat();
-    if sats > max {
-        Err(ParseError::AmountTooLarge { sats, max })
+    if sats > Amount::MAX_MONEY.to_sat() {
+        Err(ParseError::AmountTooLarge { sats })
     } else {
         Ok(())
     }
--- a/crates/btc-types/src/btc.rs
+++ b/crates/btc-types/src/btc.rs
@@ -135,66 +135,84 @@ impl<'a> Arbitrary<'a> for BitcoinOutPoint {
         Ok(BitcoinOutPoint(OutPoint { txid, vout }))
     }
 }
-/// A wrapper for bitcoin amount in sats similar to the implementation in [`bitcoin::Amount`].
+/// Validates that an amount does not exceed the bitcoin money supply
+/// ([`Amount::MAX_MONEY`]).
 ///
-/// NOTE: This wrapper has been created so that we can implement `Borsh*` traits on it.
-#[derive(
-    Arbitrary,
-    BorshSerialize,
-    BorshDeserialize,
-    Clone,
-    Copy,
-    Debug,
-    Deserialize,
-    Eq,
-    Hash,
-    Ord,
-    PartialEq,
-    PartialOrd,
-    Serialize,
-    Encode,
-    Decode,
-)]
-pub struct BitcoinAmount(u64);
-
-impl Display for BitcoinAmount {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.0)
+/// This is the single chokepoint enforced by every fallible constructor and
+/// deserialization path of [`BitcoinAmount`], so that an instance can never hold
+/// a value beyond the money supply.
+fn check_amount(sats: u64) -> Result<(), ParseError> {
+    let max = Amount::MAX_MONEY.to_sat();
+    if sats > max {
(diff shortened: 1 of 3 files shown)


crates/btc-types/src/btc.rs
```diff
@@ -135,66 +135,84 @@ impl<'a> Arbitrary<'a> for BitcoinOutPoint {
...
+    let max = Amount::MAX_MONEY.to_sat();
+    if sats > max {
+        Err(ParseError::AmountTooLarge { sats, max })
+    } else {
+        Ok(())
+    }
+}
+
+/// A wrapper around [`bitcoin::Amount`] that adds the trait impls the upstream
+/// type lacks (`Borsh*`, [`Arbitrary`], SSZ, [`Codec`]).
```
storopoli, line 153: Aren't we trying to move away from Borsh? cc @delbonis

storopoli (approved):
ACK 311d2bb18cb02c8abcfb0d1b2782367395a8aab4

crates/btc-types/src/btc.rs
prajwolrg: I think we are still using `borsh` in DB schemas. So the downstream might need it.

storopoli: We can probably remove after @delbonis finishes with https://github.com/alpenlabs/alpen/pull/1974 (no pressure 😄)

bewakes (approved):

crates/btc-types/src/btc.rs
delbonis: Leave this as-is for now.  I'll remove it when we get to it.

crates/btc-types/src/errors.rs
```diff
@@ -29,6 +29,16 @@ pub enum ParseError {
...
 
+    /// The provided amount exceeds the maximum bitcoin money supply
+    /// (`Amount::MAX_MONEY`).
+    #[error("amount of {sats} sats exceeds the maximum of {max} sats")]
+    AmountTooLarge {
+        /// The offending amount, in satoshis.
+        sats: u64,
+        /// Maximum allowed amount, in satoshis (`Amount::MAX_MONEY`).
+        max: u64,
+    },
```
delbonis: Could remove the `max` field and just have the format string directly have `{MAX_MONEY}`.

delbonis (approved):
Looks good.  One not-necessarily-actionable comment.

crates/btc-types/src/errors.rs
prajwolrg: Done in [f7bbb5c](https://github.com/alpenlabs/strata-common/pull/122/commits/f7bbb5c77dcb2eceaea7159f969d3d7b622a7b65) — dropped the `max` field and inlined `Amount::MAX_MONEY.to_sat()` in the error format string.

PR #123, opened 2026-06-18
refactor(btc-types): remove payload types

prajwolrg:
## Description

Removes the payload module (`L1Payload`, `PayloadIntent`, `PayloadSpec`, `BlobSpec`, `PayloadDest`) from `strata-btc-types`. These types describe DB-layer data, not anything the btc-types protocol layer reasons about, so they don't belong in this crate. This was flagged in #116 ("Some of these types seem like they're only used in the database, like `L1Payload`").

The upstream `L1Payload` was also semantically incorrect: it bounded each data chunk at the 520-byte Bitcoin script-element size, conflating a script push with a logical envelope payload. Chunking into script pushes is the envelope builder's job, so a checkpoint larger than 520 bytes (e.g. an epoch with account activity) could never be posted. The downstream consumer ([alpenlabs/alpen#1994](https://github.com/alpenlabs/alpen/pull/1994/commits)) has since defined its own `L1Payload`/`PayloadIntent` bounded by the envelope payload limit, leaving these upstream definitions as dead weight.

This drops the module, its SSZ delegate containers (`L1PayloadSsz`, `BlobSpecSsz`, `PayloadSpecSsz`, `PayloadIntentSsz`) and constants (`MAX_PAYLOAD_CHUNK_LEN`, `MAX_PAYLOAD_CHUNKS`, `MAX_AUX_DATA_LEN`) from `btc.ssz`, and the now-unused `strata-l1-txfmt` / `num_enum` dependencies.

### Type of Change

- [x] Refactor

## Notes to Reviewers

Downstream coordination: `alpen`'s `csm-types` still re-exports `BlobSpec`, `PayloadDest`, and `PayloadSpec` from `strata_btc_types::payload`. Once this lands and `alpen` bumps `strata-common`, those three types must be defined locally in `alpen` alongside the `L1Payload`/`PayloadIntent` it already owns (tracked by STR-3838).

## Checklist

## Related Issues

Relates to #116


storopoli (approved):
ACK 89ec0636abe1b96d6898c531240989de00940b3b

bewakes (approved):

PR #124, opened 2026-06-18
refactor(btc-types): remove GenesisL1View

prajwolrg:
## Description

Removes `GenesisL1View` from `strata-btc-types`. It was a snapshot of L1 chain state used to bootstrap header verification (block commitment, next target, epoch-start timestamp, last 11 timestamps).

That role is now served by ASM's [`L1Anchor`](https://github.com/alpenlabs/asm/blob/main/crates/btc-verification/src/header_verification.rs#L284), which carries the same anchoring fields and additionally tracks the Bitcoin `network` required to resolve consensus parameters when (re)building `HeaderVerificationState`. `GenesisL1View` has no remaining references in strata-common, alpen, or asm, so it is dead code.

`TIMESTAMPS_FOR_MEDIAN`, which lived alongside it in the same module, is retained — it is still consumed downstream by ASM header verification.

### Type of Change

- [x] Refactor

## Notes to Reviewers

Verified no usages of `GenesisL1View` across strata-common, alpen, and asm before removal.

## Checklist


storopoli (approved):
ACK 2e1edd3997512498e46a81e30be8fe9e0b92083c

bewakes (approved):

PR #143, opened 2026-07-17
fix(l1-txfmt): make MagicBytes human-readable serde round-trip-safe

prajwolrg:
## Description

`MagicBytes`'s human-readable serde was asymmetric: serialization emitted `Display`, whose non-UTF-8 fallback is the debug array string (e.g. `"[255, 254, 0, 1]"`), while deserialization required a 4-byte string and rejected that 16-character form. A `MagicBytes` built from non-UTF-8 bytes therefore serialized to JSON it could not reload, breaking round-tripping for any type that embeds it (e.g. the Alpen `BlobSpec` / consolidated params artifact).

The fix serializes the readable string only when the bytes are valid UTF-8 and falls back to the raw byte array otherwise; deserialization accepts either shape. The UTF-8 string form is unchanged, so existing artifacts keep round-tripping.

Surfaced by a Codex review on alpen PR alpenlabs/alpen#2132: https://github.com/alpenlabs/alpen/pull/2132#discussion_r3602276430

### Type of Change

## Notes to Reviewers

Reachability is low in practice: magic bytes are normally parsed from a `&str` (via `FromStr`), so they are UTF-8 by construction; the break is only reachable via `MagicBytes::new([..])` with non-UTF-8 bytes. Filed as STR-4010 (Lowest). Non-human-readable (bincode/borsh) encodings were already lossless and are untouched.

## Checklist

## Related Issues

STR-4010

### DIFF
--- a/crates/l1proto/txfmt/src/serde.rs
+++ b/crates/l1proto/txfmt/src/serde.rs
@@ -9,7 +9,15 @@ use crate::{MagicBytes, TagData};
 impl Serialize for MagicBytes {
     fn serialize<S: serde::Serializer>(&self, s: S) -> Result<S::Ok, S::Error> {
         if s.is_human_readable() {
-            s.serialize_str(&self.to_string())
+            // Prefer the readable string form, but fall back to the raw byte
+            // array when the magic is not valid UTF-8. Serializing the `Display`
+            // output for non-UTF-8 bytes yields the debug array (e.g.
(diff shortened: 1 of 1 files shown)


storopoli (approved):
ACK eeebc7c531608b57e4c7b357e1a018adfb2acd50

delbonis (approved):
Very clever!

PR #146, opened 2026-07-19
feat(logging): add owned, serde-serializable LoggingInitConfig

prajwolrg:
## Description

`strata-logging` only exposed a borrowed `LoggingInitConfig<'a>` — a struct full of `&str`/`&[&str]` meant to be assembled transiently at a call site, and additionally carrying two binary-provided constants (`service_base_name`, `default_log_prefix`) that have no business in a user-facing config file. As a result, any downstream that wants to surface logging options in a config-file section (e.g. asm-runner's `[logging]` TOML table) has to hand-roll an owned, serde-derived mirror struct plus the `Option<String>`→`Option<&str>` / `Vec<String>`→`Vec<&str>` conversion. CLI-driven consumers (alpen) sidestep this because their owned data already lives in the parsed CLI command; they build the borrowed struct transiently at the call site and never need the logging struct itself to be serde-deserializable.

This PR closes that gap by making the crate own the serde surface: a new owned `#[serde(default)]` `LoggingInitConfig` carrying only the operator-tunable subset, with `init` / `init_with_layers` methods that take the binary constants as arguments and build the borrowed view internally. The previous borrowed struct is renamed to `LoggingInitConfigRef` and remains the zero-copy view the free `init_logging_from_config*` functions consume.

Review entry point: `crates/logging/src/service.rs`.

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

The rename is source-breaking for callers that reference `LoggingInitConfig` as the *borrowed* struct (alpen's binaries, strata-bridge's common wrapper). On their next tag bump they switch either to the owned `LoggingInitConfig` (recommended for config/CLI wiring) or to `LoggingInitConfigRef` to keep the current borrowed-literal call. asm-runner is the first consumer of the owned type and will drop its local mirror once this is tagged.

Open naming question for reviewers: `LoggingInitConfigRef` now has more fields than `LoggingInitConfig` (it also holds the two binary constants), so "Ref" is a low-level-params view rather than a strict borrow of the owned type — flagging in case a different suffix reads better.

## Checklist

### DIFF
--- a/crates/logging/src/service.rs
+++ b/crates/logging/src/service.rs
@@ -2,13 +2,93 @@
 
 use std::path::PathBuf;
 
+use serde::{Deserialize, Serialize};
 use tracing::info;
 
 use super::{BoxedLayer, FileLoggingConfig, LoggerConfig, format_service_name, init_with_layers};
 
-/// Configuration parameters for logging initialization.
+/// Owned, serde-serializable logging configuration.
+///
+/// This is the type binaries should embed in their own config structs (e.g. a
+/// `[logging]` TOML section) or populate from CLI flags. Every field is
+/// optional, so a partial or omitted section deserializes cleanly and falls
+/// back to the crate defaults.
+///
+/// It carries only the operator-tunable subset of the settings. The two
+/// binary-provided constants — the base service name and the default log-file
+/// prefix — are passed to [`init`](Self::init) / [`init_with_layers`](Self::init_with_layers)
+/// as arguments rather than serialized, so they never leak into a user's config
+/// file.
+///
+/// Under the hood these methods build a [`LoggingInitConfigRef`] (the borrowed,
+/// zero-copy view the init routines consume) and forward to
+/// [`init_logging_from_config`] / [`init_logging_from_config_with_layers`].
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(default)]
+pub struct LoggingInitConfig {
+    /// Optional service label appended to the service name (e.g. `"prod"`, `"dev"`).
+    pub service_label: Option<String>,
+    /// OpenTelemetry OTLP collector endpoint. When set, OTLP export is enabled.
+    pub otlp_url: Option<String>,
+    /// Directory to write rolling log files into. When unset, file logging is disabled.
+    pub log_dir: Option<PathBuf>,
+    /// Filename prefix for rolling log files. Falls back to the binary's default prefix.
+    pub log_file_prefix: Option<String>,
+    /// Use JSON output format instead of the compact text format.
+    pub json_format: Option<bool>,
+    /// Extra `EnvFilter` directives applied before `RUST_LOG` (e.g. to silence
+    /// noisy dependencies). Empty when omitted.
+    pub extra_filter_directives: Vec<String>,
+}
+
+impl LoggingInitConfig {
+    /// Initialize process-global logging from this config.
+    ///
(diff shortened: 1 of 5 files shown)


crates/logging/src/service.rs
```diff
@@ -2,13 +2,93 @@
 
 use std::path::PathBuf;
 
+use serde::{Deserialize, Serialize};
 use tracing::info;
 
 use super::{BoxedLayer, FileLoggingConfig, LoggerConfig, format_service_name, init_with_layers};
 
-/// Configuration parameters for logging initialization.
+/// Owned, serde-serializable logging configuration.
```
storopoli: Why do we need "Owned" here?


```suggestion
/// Configuration parameters for logging initialization.
```

crates/logging/src/service.rs
```diff
@@ -2,13 +2,93 @@
...
+                default_log_prefix,
+                extra_filter_directives: &extra_filter_directives,
+            },
+            extra_layers,
+        );
+    }
+}
+
+/// Borrowed, zero-copy view of the parameters [`init_logging_from_config`]
+/// consumes.
```
storopoli: Same here

```suggestion
/// Zero-copy view of [`LoggingInitConfig`].
```

storopoli (changes requested):
Docstrings are confusing for downstream implementers which should be the primary target audience.

crates/logging/src/service.rs
prajwolrg: Applied the suggestion in 381796804c5ca3234f3bb24f8b10873ba840f73e, along with dropping the same ownership jargon from the other paragraphs (the "Under the hood … borrowed, zero-copy view" internals bit is gone). The docstrings were then reworked more substantially in 6b16353f77cd1a1dd7d97836586e487ab840c6de to describe observable behavior and defaults for downstream implementers, with a TOML + `init` usage example.

crates/logging/src/service.rs
prajwolrg: Applied in 381796804c5ca3234f3bb24f8b10873ba840f73e — summary line now matches the suggestion, and the "Prefer the owned … owning struct" paragraph below it was rephrased to plain when-to-use guidance.

crates/logging/src/service.rs
```diff
@@ -2,13 +2,88 @@
...
+/// [`init_with_layers`](Self::init_with_layers). Every field is optional, so a
+/// partial or omitted section deserializes cleanly and falls back to the crate
+/// defaults.
+///
+/// It carries only the operator-tunable subset of the settings. The two
+/// binary-provided constants — the base service name and the default log-file
+/// prefix — are passed to the init methods as arguments rather than
+/// serialized, so they never leak into a user's config file.
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[serde(default)]
```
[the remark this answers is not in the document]
prajwolrg: Good catch. Addressed in 5ccda4a1cd9511f221abe1907cd3f74760e1e683: the struct docs now state that omitting the whole section requires `#[serde(default)]` on the embedding field (the compile-checked doctest example shows exactly that pattern), and a new test pins both behaviors — an embedding with the field-level default accepts `{}`, and a bare embedding without it rejects the same input. An embedding mechanism that supplies the default from the inner type is not possible with serde, so documenting + testing the field attribute is the fix.

voidash (approved):

storopoli (approved):
ACK 5ccda4a1cd9511f221abe1907cd3f74760e1e683

PR #148, opened 2026-07-22
STR 4058 Shared commit-reveal builder and parser

krsnapaudel:
## Description

Adds shared SPS-53 commit/reveal format support to `strata-common`.

- **`strata-l1-envelope-fmt`**: adds a strict signed-envelope leaf parser (`parse_signed_envelope_leaf` → `SignedEnvelopeLeaf`) and a matching builder. The lenient parsers (`parse_envelope_payload`, `parse_multi_envelope_payloads`) remain for scan-forward, data-carrying scripts.
- **`strata-l1-commit-reveal-fmt`** (new crate): SPS-53 commit/reveal script building (`OP_RETURN` marker + one signed reveal leaf per chunk) and reading — single-set payload extraction for the closed-witness (guest) path, plus stateless, consumer-neutral **grouping helpers** for host scanners: classify a commit candidate, assign a reveal to the commit whose slot it spends, and extract a grouped commit's payload. The caller owns block iteration, accumulation, reorg, and telemetry; marker-tail (version) and key checks stay with the caller.
- **`test-utils`**: optional fixtures for downstream crates that need commit/reveal transaction shapes in tests.

The strict parser accepts only `<pubkey> OP_CHECKSIG OP_FALSE OP_IF <payload pushes> OP_ENDIF`, so a confirmed reveal spend is evidence the pubkey holder signed — no later opcode can invert the `OP_CHECKSIG` result.

### Breaking changes (`strata-l1-envelope-fmt`)

- `EnvelopeParseError::MissingChecksigverify` renamed to `MissingChecksig`.
- `parse_envelope_container` now enforces the strict single-leaf grammar (delegates to `parse_signed_envelope_leaf`); callers needing lenient extraction should use `parse_envelope_payload` / `parse_multi_envelope_payloads`.

This PR was created with help from Claude and Codex.

### Testing
Tested the use of the shared crate in `alpen`: `alpen-ee/da/l1-extraction` (WIP), `btcio/src/writer/chunked_envelope/builder` and `alpen-ee/da/runtime/src/verification/`. WIP in `STR-4059` branch of `alpen`.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers
The commit-reveal builder keeps two entry points on purpose: `build_commit_reveal_scripts` (splits a payload into chunks at build time) and `build_commit_reveal_scripts_from_chunks` (one reveal leaf per supplied chunk, no re-split). The EE DA provider in `alpen` persists `ChunkedEnvelopeEntry`, which the L1 writer reads. It must build reveal leaves using the stored chunks.

## Checklist

## Related Issues

[STR-4058](https://alpenlabs.atlassian.net/browse/STR-4058)

[STR-4058]: https://alpenlabs.atlassian.net/browse/STR-4058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/l1proto/commit-reveal/src/builder.rs
+++ b/crates/l1proto/commit-reveal/src/builder.rs
@@ -1,0 +1,363 @@
+//! Script construction for one commit-reveal set.
+
+use std::num::NonZeroUsize;
+
+use bitcoin::blockdata::script::Builder;
+use bitcoin::opcodes::all::OP_RETURN;
+use bitcoin::script::PushBytesBuf;
+use bitcoin::{Script, ScriptBuf};
+use strata_l1_envelope_fmt::SIGNED_LEAF_PUBKEY_LEN;
+use strata_l1_envelope_fmt::builder::{
+    build_signed_envelope_leaf, split_payload_into_envelope_chunks,
+};
+use strata_l1_txfmt::MagicBytes;
+
+use crate::MAX_MARKER_TAIL_BYTES;
+use crate::errors::CommitRevealBuildError;
+
+/// The scripts for one commit-reveal set.
+///
+/// Holding this is the writer-side contract: if it carries N reveal leaves, the
+/// commit transaction must fund exactly N reveal-slot outputs.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CommitRevealScripts {
+    marker_script: ScriptBuf,
+    reveal_leaf_scripts: Vec<ScriptBuf>,
+}
+
+impl CommitRevealScripts {
+    /// The commit marker script.
+    pub fn marker_script(&self) -> &Script {
+        &self.marker_script
+    }
+
+    /// The reveal leaf scripts, in commit-output order.
+    pub fn reveal_leaf_scripts(&self) -> &[ScriptBuf] {
+        &self.reveal_leaf_scripts
+    }
+
+    /// How many P2TR reveal slots the commit transaction must fund.
+    ///
+    /// Never zero: an empty chunk list is rejected with
+    /// [`CommitRevealBuildError::NoChunks`].
+    pub fn reveal_slot_count(&self) -> NonZeroUsize {
+        NonZeroUsize::new(self.reveal_leaf_scripts.len())
+            .expect("builder rejects an empty chunk list, so there is at least one leaf")
+    }
+
+    /// Consumes the set, returning the marker script and the leaf scripts.
+    pub fn into_parts(self) -> (ScriptBuf, Vec<ScriptBuf>) {
+        (self.marker_script, self.reveal_leaf_scripts)
+    }
+}
+
+/// Builds the marker and one reveal leaf per supplied chunk.
+///
+/// `marker_tail` follows the magic in the marker and is not interpreted here.
+/// Chunk boundaries are preserved: each supplied chunk becomes exactly one
+/// reveal leaf.
+///
+/// # Errors
+///
+/// [`CommitRevealBuildError::NoChunks`] if `chunks` is empty,
+/// [`CommitRevealBuildError::EmptyChunk`] if any chunk is empty,
+/// [`CommitRevealBuildError::MarkerTailTooLarge`] if `marker_tail` exceeds
+/// [`MAX_MARKER_TAIL_BYTES`], or [`CommitRevealBuildError::RevealLeaf`] if a leaf
+/// cannot be built.
+pub fn build_commit_reveal_scripts_from_chunks(
+    magic: &MagicBytes,
+    marker_tail: impl AsRef<[u8]>,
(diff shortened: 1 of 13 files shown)


crates/l1proto/envelope-fmt/src/parser.rs
```diff
@@ -101,6 +105,156 @@ pub fn parse_envelope_container(
...
+///
+/// The pubkey must be exactly [`SIGNED_LEAF_PUBKEY_LEN`] bytes. See
+/// [`EnvelopeParseError::InvalidPubkeyLength`] for why a laxer rule would make
+/// the authentication meaningless.
+///
+/// # Errors
+///
+/// Returns [`EnvelopeParseError`] if the script deviates from the shape above
+/// in any way, or if the payload exceeds [`MAX_ENVELOPE_PAYLOAD_SIZE`].
+pub fn parse_exact_signed_envelope_leaf(
```
prajwolrg: I don't think this is needed. We do have `parse_envelope_container` that extracts the pubkey. Maybe we can modify it to handle this. If we strictly require something, I think the consumer should handle that. A lot of the functionality here feels duplicated.

prajwolrg (changes requested):
This is a good direction. It looks like there are a lot of duplication to the existing code. I think we should be able to use a lot of existing functions here.

crates/l1proto/envelope-fmt/src/parser.rs
```diff
@@ -82,9 +86,9 @@ pub fn parse_envelope_container(
         _ => return Err(EnvelopeParseError::MissingPubkey),
     };
 
-    // Verify CHECKSIGVERIFY
+    // Verify OP_CHECKSIG
     if next_op(&mut instructions) != Some(OP_CHECKSIG) {
-        return Err(EnvelopeParseError::MissingChecksigverify);
+        return Err(EnvelopeParseError::MissingChecksig);
```
prajwolrg: Thanks for correcting this.

crates/l1proto/commit-reveal/src/lib.rs
```diff
@@ -0,0 +1,65 @@
...
+//! Where [`strata_l1_envelope_fmt`] handles one script envelope, this crate
+//! handles the multi-transaction carrier built on top of it: a commit
+//! transaction whose outputs fund N reveal transactions, each carrying one
+//! chunk of a larger payload.
+//!
+//! ```text
+//! commit tx:
+//!   vout 0    = OP_RETURN <single push: magic || consumer-defined tail>
+//!   vout 1..N = P2TR reveal slots (contiguous run)
+//!   change afterwards, which MUST NOT be P2TR
```
prajwolrg, line 11: I'm curious why this must not be P2TR.

krsnapaudel: @prajwolrg I think this just a marker, per spec, to signal the end of reveals, since all reveals are P2TR.

crates/l1proto/commit-reveal/src/builder.rs
```diff
@@ -0,0 +1,378 @@
...
+    /// Returns [`CommitRevealBuildError::NoChunks`] if `chunks` is empty, since
+    /// a commit with no reveal slots cannot be read back. Returns other
+    /// variants if the marker exceeds [`MAX_MARKER_PAYLOAD_BYTES`] or a leaf
+    /// script cannot be built.
+    pub fn build_scripts(
+        &self,
+        marker_tail: &[u8],
+        reveal_pubkey: &[u8; SIGNED_LEAF_PUBKEY_LEN],
+        chunks: &[impl AsRef<[u8]>],
+    ) -> Result<CommitRevealScripts, CommitRevealBuildError> {
```
delbonis: This builder is completely useless.  The only thing it exists to do is to make it more annoying to call `CommitRevealBuilder::build_scripts` by making the caller have to construct a builder first to call it instead of just being able to pass the magic value into a standalone fn directly.

It does not:

* handle chunking a large message into multiple chunks
* handle splitting a large message over multiple reveal transactions
* handle signing or key management
* improve ergonomics of the "marker tail"

So what purpose does it serve?

delbonis (commented):
Will DM.

crates/l1proto/envelope-fmt/src/parser.rs
krsnapaudel: @prajwolrg I am addressing the duplication.

crates/l1proto/commit-reveal/src/builder.rs
krsnapaudel: @delbonis I think signing does not belong here. Correct me if I am wrong. I have addressed the others.

PR #152, opened 2026-07-29
feat(ssz): add delegate ref view for generated schemas

krsnapaudel:
## Description

Types using `impl_ssz_via_delegate!` can't be referenced as external containers in `ssz-gen` schemas — `external_kind: container` resolves to a `{Type}Ref<'a>` they don't have, so consumers hand-write one. See [`asm`](https://github.com/alpenlabs/asm/blob/main/crates/btc-verification/src/ssz_delegate.rs) for `HeaderVerificationState`.

Adds `SszDelegateRef<'a, T>` to `strata-identifiers`. It borrows the delegate's own view — named through the delegate's `SszHasView` link, so it can't be named wrongly — and converts on materialization, letting `from_delegate`'s error surface as a `DecodeError` instead of a panic.

A wrapper declares `SszDelegate` and one alias:

```rust
pub type HeaderVerificationStateRef<'a> = SszDelegateRef<'a, HeaderVerificationState>;
```

This PR was created with help from Claude and Codex.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers
Need to update the `ssz-gen` deps after [`ssz-gen#101`](https://github.com/alpenlabs/ssz-gen/pull/101) goes in.

## Checklist

## Related Issues

[STR-3948](https://alpenlabs.atlassian.net/browse/STR-3948)

[STR-3948]: https://alpenlabs.atlassian.net/browse/STR-3948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/crates/identifiers/src/macros/ssz.rs
+++ b/crates/identifiers/src/macros/ssz.rs
@@ -1,26 +1,19 @@
-//! SSZ view trait macros for identifier and wrapper types.
+//! SSZ implementations for hand-written types used in ssz-gen schemas.
 //!
-//! These macros generate the boilerplate trait implementations (`DecodeView`,
-//! `SszTypeInfo`, `TreeHash`, `ToOwnedSsz`) that the SSZ view layer requires.
-//! Three macros are provided, each targeting a different structural pattern:
+//! ssz-gen emits `Encode`/`Decode`/`TreeHash` and a byte-backed `{Type}Ref` view
+//! only for types a schema declares. Types written in Rust get theirs from here.
 //!
-//! | Macro | Use when… |
-//! |---|---|
-//! | [`impl_ssz_fixed_container!`] | Multi-field struct with `#[ssz(struct_behaviour = "container")]` |
-//! | [`impl_ssz_transparent_wrapper!`] | Newtype whose inner type already implements `DecodeView` |
-//! | [`impl_ssz_transparent_byte_array_wrapper!`] | Newtype wrapping a raw `[u8; N]` (which lacks `DecodeView`) |
+//! ## Giving a type its SSZ impls
 //!
-//! ## Choosing the right macro
+//! - [`impl_ssz_fixed_container!`] — multi-field struct with `#[ssz(struct_behaviour =
+//!   "container")]`, all fields fixed-size.
+//! - [`impl_ssz_transparent_wrapper!`] — newtype whose inner type already implements `DecodeView`.
+//! - [`impl_ssz_transparent_byte_array_wrapper!`] — newtype wrapping a raw `[u8; N]`.
+//! - [`impl_ssz_via_delegate!`] — wrapper whose encoding is another SSZ type's, declared with
+//!   [`SszDelegate`].
 //!
-//! ```text
-//!  Is the type a multi-field container?
-//!    ├─ Yes → impl_ssz_fixed_container!
-//!    └─ No (newtype / transparent wrapper)
-//!         ├─ Inner type has DecodeView? (Buf32, RBuf32, u64, …)
-//!         │    └─ Yes → impl_ssz_transparent_wrapper!
-//!         └─ Inner type is [u8; N]?
-//!              └─ Yes → impl_ssz_transparent_byte_array_wrapper!
-//! ```
+//! Exactly one applies to a given type: they all emit `SszTypeInfo`/`TreeHash`,
+//! so combining two is a coherence error.
 //!
 //! The split between the two transparent-wrapper macros exists because `[u8; N]`
 //! does **not** implement `DecodeView` in the `ssz` crate — only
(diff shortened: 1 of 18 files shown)


crates/identifiers/src/macros/ssz.rs
```diff
@@ -269,6 +269,85 @@ pub trait SszDelegate: Sized {
...
+/// [`ToOwnedSsz::to_owned`](ssz_types::view::ToOwnedSsz::to_owned) is not, so
+/// the conversion has to happen during decoding.
+///
+/// A crate exposes this under the name a generated schema expects by aliasing
+/// it, e.g. `pub type FooRef<'a> = SszDelegateRef<'a, Foo>;` in the crate that
+/// defines `Foo`.
+// Not `Copy`: it would overlap `ssz_types`' `impl<T: Copy> ToOwnedSsz<T> for T`.
+#[derive(Clone, Debug)]
+pub struct SszDelegateRef<'a, T> {
+    inner: T,
```
delbonis: This is unfortunate.  It kinda defeats the purpose of view types since it forces a bunch of copies.  It's not *really* an "eager SSZ view", it's a fake view.  What else can we do here?

crates/ssz-delegate-tests/ssz/delegate.ssz
delbonis, line 1: Is it possible to add this to the `tests` dir inside some crate instead of adding a new crate for it?

delbonis (changes requested):

crates/identifiers/src/macros/ssz.rs
krsnapaudel: @delbonis You're right. We could update the doc to be honest or implement the delegate-backed view as `SszDelegateRef<'a, T>`. `HeaderVerificationState` can't use it since `from_delegate` rejects unknown network ids and out-of-bounds ring-buffer heads, and `ToOwnedSsz::to_owned` returns `T`, not `Result<T>`. A borrowing view would have to panic on those bytes, or hold unvalidated bytes while claiming to be a validated `HeaderVerificationState`.

krsnapaudel: @delbonis Please check commit 2 for an alternative. It seems this is better than commit 1.

crates/ssz-tests/build.rs
```diff
@@ -0,0 +1,27 @@
...
+
+fn main() {
+    let out_dir = std::env::var("OUT_DIR").expect("OUT_DIR not set by cargo");
+    let output_path = Path::new(&out_dir).join("generated_delegate_ssz.rs");
+
+    let entry_points = ["delegate.ssz"];
+    // No external crates: the schema's `import consumer` resolves to the
+    // integration test's own `consumer` module, which declares the wrappers and
+    // their view aliases.
+    let crates: [&str; 0] = [];
```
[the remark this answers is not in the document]
krsnapaudel: It seems this makes no difference for tests.

crates/identifiers/src/macros/ssz.rs
```diff
@@ -269,18 +277,270 @@ pub trait SszDelegate: Sized {
...
 }
 
+/// A delegate wrapper whose conversion from its delegate cannot fail.
+///
+/// Every value of [`SszDelegate::Delegate`] must be a valid `Self`. Wrappers
+/// that reject some delegate values must not implement it.
+pub trait SszDelegateInfallible: SszDelegate {
+    /// Reconstructs `Self` from a delegate that is valid by construction.
+    fn from_delegate_infallible(delegate: Self::Delegate) -> Self;
+}
```
delbonis: Should have a `impl<T: SszDelegateInfallible> SszDelegate for T` that calls this.  If there's an infallible impl then that naturally implies an impl for the looser type.

However, this is actually a red herring, the issue is larger in scope, see main comment.

delbonis (changes requested):
There's a lot of repetitiveness here that I'm not really comfortable with.  It might be prompting a combinatorial blowup of the SSZ variants with this approach.

**But there's a bigger issue:** The point of the `SszDelegate` idea is to allow for a slight separation of syntax and semantics of SSZ types, where we define a "first pass" using the pythonic defs and then can add some structural checks to it when converting it from the *delegatee* to the real SSZ type we want, which *can fail* but get bubbled into the native SSZ decode error variants (as a custom error).  So having the "infallible" variants kinda defeats the purpose, because if the conversion can't fail then there's no reason to have the conversion step as opposed to decoding directly to the type.

Supporting this SSZ delegate pattern with the ref types *should* be possible, although I haven't thought deeply about how.  But the above issue defeats the point of the whole pattern, so this is definitely an incorrect approach.

delbonis:
This PR should probably be rethought in a followup.  Closing now, can re-do after some design work.

PR #154, opened 2026-07-31
Array-based identifier trait

delbonis:
## Description

This PR adds a trait to capture identifiers that are isomorphic to fixed-size byte arrays of some size, and then implements it on several types.  This seems like it'd be useful for making the database layer more generic in the future.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

I wish we had magical future Rust features so this could be a lot cooler!  This is kinda limited as-is, unfortunately.

## Checklist

### DIFF
--- a/crates/identifiers/src/array_keys/macros.rs
+++ b/crates/identifiers/src/array_keys/macros.rs
@@ -1,108 +1,7 @@
-//! Utilities for working with identifiers as database keys.
+//! Macros for declaring [`ArrayKey`] impls, along with the impls we declare
+//! with them for the primitive types.
 
-/// Used for types that can be represented as a flat array of bytes, as would be
-/// used for binary database keys.
-///
-/// Unfortunately, `generic_const_exprs` is not stable, so we can't make this
-/// interface *really* nice, we just have to hope that LLVM realizes that it can
-/// eliminate all these bounds checks.
-pub trait ArrayKey {
-    /// The size of an array that contains this key byte.
-    const BYTES: usize;
-
-    /// Copies the key bytes into the array.
-    ///
-    /// This must be big-endian.
-    ///
-    /// # Panics
-    ///
-    /// If called with a slice that is not [`Self::BYTES`] bytes.
-    fn copy_into(&self, buf: &mut [u8]);
-
-    /// Constructs an instance of the key by decoding from bytes.
-    ///
-    /// # Panics
-    ///
-    /// If called with a slice that is not [`Self::BYTES`] bytes.
-    fn copy_from(buf: &[u8]) -> Self;
-}
-
-/// General impl for all bytebufs since they're already of the right format.
-impl<const N: usize> ArrayKey for [u8; N] {
-    const BYTES: usize = N;
-
-    fn copy_into(&self, buf: &mut [u8]) {
-        assert_buf_len_eq::<Self>(buf);
-        buf.copy_from_slice(self);
-    }
-
-    fn copy_from(buf: &[u8]) -> Self {
-        assert_buf_len_eq::<Self>(buf);
-        buf.try_into().unwrap()
-    }
-}
-
-/// Extension trait for "buffer types" that we might want to decode as an [`ArrayKey`].
-pub trait ArrayKeyBuf {
-    /// Attempts to decode the buf into an [`ArrayKey`] type.
-    ///
-    /// Returns `None` if the array is not the right size.
-    fn try_into_key<K: ArrayKey>(&self) -> Option<K>;
-}
-
-impl<T: AsRef<[u8]>> ArrayKeyBuf for T {
-    fn try_into_key<K: ArrayKey>(&self) -> Option<K> {
-        let buf = <Self as AsRef<[u8]>>::as_ref(self);
-        if buf.len() == K::BYTES {
-            Some(K::copy_from(buf))
-        } else {
-            None
-        }
-    }
-}
-
-/// Extension trait for array types that we might want to decode as an [`ArrayKey`].
-///
-/// Unlike [`ArrayKeyBuf`], this is infallible, since the array length is known
-/// statically and MUST be checked at compile time.
-///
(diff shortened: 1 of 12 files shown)


crates/identifiers/src/key_types.rs
```diff
@@ -0,0 +1,371 @@
...
+    };
+}
+
+pub(crate) use decl_array_key_struct_impl;
+pub(crate) use decl_array_key_wrapper_impl;
+
+#[inline(always)]
+pub(crate) fn assert_buf_len_eq<K: ArrayKey>(arr: &[u8]) {
+    #[cfg(debug_assertions)]
+    {
```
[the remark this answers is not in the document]
delbonis: Shit maybe this is a good point.

crates/identifiers/src/array_keys/macros.rs
```diff
@@ -0,0 +1,330 @@
...
+    };
+}
+
+decl_array_key_tuple_impl!(A, B);
+decl_array_key_tuple_impl!(A, B, C);
+decl_array_key_tuple_impl!(A, B, C, D);
+decl_array_key_tuple_impl!(A, B, C, D, E);
+decl_array_key_tuple_impl!(A, B, C, D, E, F);
+
+/// Declares an [`ArrayKey`] impl for an integer type, encoding it big-endian.
```
storopoli, line 53: Will we ever need a little-endian version?
If so then maybe this macro should be prefixed `_be`.

crates/identifiers/src/array_keys/macros.rs
```diff
@@ -0,0 +1,330 @@
...
+            const BYTES: usize = core::mem::size_of::<$ty>();
+
+            fn copy_into(&self, buf: &mut [u8]) {
+                assert_buf_len_eq::<Self>(buf);
+                buf.copy_from_slice(&<$ty>::to_be_bytes(*self)[..]);
+            }
+
+            fn copy_from(buf: &[u8]) -> Self {
+                assert_buf_len_eq::<Self>(buf);
+                <$ty>::from_be_bytes(buf.try_into().unwrap())
```
storopoli, line 66: Are we fine with this `unwrap()`? Why not use an `expect("foo")` here?

crates/identifiers/src/array_keys/macros.rs
```diff
@@ -0,0 +1,330 @@
...
+        }
+    }
+
+    /// Deterministic pseudorandom byte source so that failures are reproducible.
+    fn gen_entropy(seed: u64, len: usize) -> Vec<u8> {
+        let mut state = seed | 1;
+        (0..len)
+            .map(|_| {
+                state ^= state << 13;
+                state ^= state >> 7;
```
storopoli, line 177: Why not just derive `Arbitrary` and use `proptest!` for these?

crates/identifiers/src/array_keys/traits.rs
```diff
@@ -0,0 +1,124 @@
...
+///
+/// You shouldn't have to implement this on your own types, it's just so we can
+/// add trait member fn to `[u8; N]`.
+pub trait ArrayKeyArr<const N: usize> {
+    /// Decodes the array into an [`ArrayKey`] type of exactly the same width.
+    ///
+    /// Instantiating this with a key type whose width isn't `N` MUST fail to
+    /// compile.
+    fn into_key<K: ArrayKey>(self) -> K;
+}
```
storopoli, line 77: Can't we hide this away as a sealed trait somewhere or even not make this `pub`?
I don't like docstrings pinky swear guarantees like:

> You shouldn't have to implement this on your own types...

storopoli (approved):
Really nice. Couple non-blocking comments.

ACK 5ade297da2e4bc8a50ed5836ddf8c2c98fe0c197

crates/identifiers/src/array_keys/macros.rs
delbonis: Actually it's undesirable to support little endian because it breaks sorting, so that should never be a feature.

crates/identifiers/src/array_keys/macros.rs
delbonis: The fn is defined (in the trait) to panic if the buf is the wrong length and we already panic in the `assert_buf_len_eq` call, with a richer message, so this unwrap would never actually trigger.

Consumer should *try* to interact with these types through the `ArrayKeyBuf`/`ArrayKeyArr` traits instead of the `copy_into`/`copy_from` fns.

crates/identifiers/src/array_keys/macros.rs
delbonis: This is all Claude generated, will resolve.

Not sure how extensive it makes sense to make this since there's no data-dependent branches aside from bounds checks.

crates/identifiers/src/array_keys/traits.rs
delbonis: All of this stuff is because we can't make args dependent on trait consts because those are nightly features.

I really wanted to have something like `pub type KeyBuf<A: ArrayKey> = [u8; A::BYTES];` but even that's not allowed.

crates/identifiers/src/array_keys/macros.rs
storopoli: Ok, solving.

crates/identifiers/src/array_keys/traits.rs
storopoli: What if we use the nightly feature. I think it's const generic right?
We are already using nightly here and in the downstream deps: `moho`, `asm`, `alpen` `alpen-ee` etc...

delbonis: Ideally we wouldn't be using the nightly compiler in proofs because we don't want subtle codegen bugs ending up causing our proofs to be unsound.

PR #155, opened 2026-08-01
fix(predicate): stop panicking on untrusted predicate key input

prajwolrg:
## Description

`strata-predicate` panicked instead of returning an error in two places reachable from untrusted input:

- `PredicateKey::new` converted the caller-supplied condition bytes into a length-bounded SSZ list with `.expect("valid length")`, and `PredicateKeyBuf`'s `TryFrom<&[u8]>` never checked the condition length at all. Both are reachable from deserialization (serde JSON/bincode, `arbitrary`, and borsh via `PredicateKeyBuf::try_from` + `to_owned`), so an oversized condition in externally supplied bytes crashed the process instead of failing to decode.
- `PredicateKey::as_buf_ref` panicked whenever the raw SSZ-decoded type id wasn't a registered `PredicateTypeId`. That's not a can't-happen case: the id is stored as an unvalidated raw byte specifically so the format can round-trip predicate types a given build doesn't know about. `try_as_buf_ref` already existed to handle this gracefully.

The constructor is renamed `try_new` and returns `Result`; `as_buf_ref` is removed in favor of the existing `try_as_buf_ref`.

### Type of Change

- [x] New or updated tests

## Notes to Reviewers

This is a public API break for `PredicateKey::new` (now `try_new`, returns `Result`) and removes `PredicateKey::as_buf_ref`. Downstream repos (alpen, alpen-ee, asm, strata-bridge) pin this crate via git tag/rev, so nothing breaks until they bump the dependency; call sites there will need updating separately when that happens.

## Checklist

## Related Issues

N/A

### DIFF
--- a/crates/predicate/src/errors.rs
+++ b/crates/predicate/src/errors.rs
@@ -21,6 +21,15 @@ pub enum PredicateError {
     #[error("missing predicate type")]
     MissingPredicateType,
 
+    /// Predicate condition bytes exceed the maximum allowed length.
+    #[error("condition length {len} exceeds maximum of {max} bytes")]
+    ConditionTooLong {
+        /// The length of the provided condition bytes.
+        len: usize,
+        /// The maximum allowed condition length.
+        max: usize,
+    },
+
     /// Unknown predicate type name in string parsing.
     #[error("unknown predicate type name: {0}")]
     UnknownPredicateTypeName(String),
--- a/crates/predicate/src/errors.rs
+++ b/crates/predicate/src/errors.rs
@@ -7,6 +7,7 @@
 
 use thiserror::Error;
 
+use crate::MAX_CONDITION_LEN;
 use crate::type_ids::PredicateTypeId;
 
 /// Errors that can occur when working with predicates, claims, and witnesses.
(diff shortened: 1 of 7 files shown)


storopoli (approved):
ACK 3f7df2f3e6c92ef91ad063ed3e48a271cbe031c9

crates/predicate/src/errors.rs
```diff
@@ -21,6 +21,15 @@ pub enum PredicateError {
     #[error("missing predicate type")]
     MissingPredicateType,
 
+    /// Predicate condition bytes exceed the maximum allowed length.
+    #[error("condition length {len} exceeds maximum of {max} bytes")]
+    ConditionTooLong {
```
delbonis, line 26: Since it's defined as a constant we could interpolate the constant value in the message instead of including it as a variant field.

delbonis (approved):
Minor comment, could ignore.

crates/predicate/src/errors.rs
prajwolrg: Makes sense. Fixed in [ce5556c](https://github.com/alpenlabs/strata-common/pull/155/commits/ce5556cd4ffa3cda58393e36855b092714971ac8).

PR #156, opened 2026-08-01
chore(deps): fix cargo audit vulnerability (RUSTSEC-2026-0220)

prajwolrg:
## Description

`cargo audit` was failing CI's supply-chain job with a vulnerability in `ruint` 1.19.0 (RUSTSEC-2026-0220: shift operations produce incorrect overflow flags and truncated shift amounts). It's pulled in transitively through `ssz_primitives` (from ssz-gen). The fix version is within the existing semver range accepted by its dependents, so `cargo update -p ruint` resolves it to 1.20.0 without needing to touch the ssz-gen pin.

`cargo audit` now exits clean apart from the four pre-existing unmaintained/unsound advisories (`ansi_term`, `bincode`, `paste`, `scc`) that the CI job already treats as non-blocking (`cargo audit -D warnings` runs with `continue-on-error: true`); only the plain `cargo audit` vulnerability check gates the job, and that now passes.

### Type of Change

- [x] Dependency update
- [x] Security fix

## Notes to Reviewers

Scope is intentionally limited to the one advisory that actually fails CI. The lockfile diff also flips `winapi-util`'s `windows-sys` dependency edge between two versions that were already both present in the lockfile (0.48.0 and 0.61.2) — that's the resolver re-verifying after `ruint` was unlocked, not an intentional change; confirmed a plain `cargo check` on the prior lockfile doesn't touch it.

## Checklist

## Related Issues

N/A

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2630,9 +2630,9 @@ dependencies = [
 
 [[package]]
 name = "ruint"
-version = "1.19.0"
+version = "1.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45caf26f647c19115bf9c453c70ffe4a4a3a6390dceebd942610584f99b8ddce"
+checksum = "f5e99bff0393163bb25029a6af25d3d8d202ba5b5438a74d1bd8789f5c822970"
 dependencies = [
  "proptest",
  "rand 0.8.7",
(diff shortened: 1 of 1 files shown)


storopoli (approved):
ACK 8dce710eada57346653581d836b04ec577412695

PR #167, opened 2026-08-07
STR 4058 Shared commit-reveal parser

krsnapaudel:
## Description

Add shared payload parsing for signed-envelope carriers.

Extends `strata-l1-envelope-fmt` with shared parsing for the strict SPS-51 signed envelope leaf and marker-anchored payload carriers. The parser reads marker payloads as `magic || opaque tail`, extracts producer public keys from strict signed reveal leaves and preserves ordered payload chunks without forcing concatenation. It reports malformed protocol transactions with typed errors. Unrelated transactions and reveals commit after anchoring commits are ignored.

For SPS-53 chunked commit-reveal carriers, the parser derives reveal-slot ranges from the contiguous P2TR output run, matches reveal transactions to pending anchors, rejects malformed reveal sets, and prunes incomplete state by L1 height.

API shape:
- `PayloadParserConfig::single_reveal(magic)`: for single-transaction carriers such as checkpoint parsing, where the marker and signed payload leaf are on the same transaction.
- `PayloadParserConfig::chunked_reveals(magic)`: for EE DA/SPS-53 scanning, where one marker-bearing commit anchors one or more reveal transactions.
- `PayloadParser::parse(txs, l1_block_height)`: accepts `Transaction` items (and block height) and returns completed `RecoveredPayload`s. Errors are logged using `warn!`. For chunked carriers it retains incomplete payloads across calls.
- `RecoveredPayload`: common output with `anchor_txid`, opaque `tail`, `producer_pubkey`, borrowed `chunks()`, and consuming `into_chunks()` / `into_payload()` helpers.
- `parse_signed_envelope_leaf`: strict SPS-51 leaf parser for callers that need the lower-level leaf primitive directly.

### Type of Change

- [ ] Documentation update
- [x] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers
1. The shared parser is called `PayloadParser` and its entry point is `parse`. Would appreciate better name suggestions.
2. Existing low-level api has not be changed to `pub(crate)`.
3. Multiple envelopes within a single leaf is not supported although a low level api exists (not added by this PR).
4. Pruning is done inside `parse`. Would appreciates suggestions for something better.

## Checklist

## Related Issues

[STR-4058](https://alpenlabs.atlassian.net/browse/STR-4058)

[STR-4058]: https://alpenlabs.atlassian.net/browse/STR-4058?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


crates/l1proto/envelope-fmt/src/parser/payload.rs
```diff
@@ -0,0 +1,895 @@
...
+    }
+
+    /// Parses transactions and returns recovered payloads and rejected transactions.
+    ///
+    /// Errors encountered while parsing protocol-relevant commit and reveals
+    /// are included as `error` inside [`RejectedTx`].
+    pub fn parse<'a>(
+        &mut self,
+        txs: impl IntoIterator<Item = L1IncludedTxRef<'a>>,
+    ) -> PayloadParserOutput {
```
delbonis: This signature is weird.  The height is passed with each tx, which would allow us to pass transactions from interleaved heights in the same pass.  There is never a scenario where we'd want to do that.  This should be an iterator of `&Transaction` and then pass the height as an argument.

crates/l1proto/envelope-fmt/src/parser/payload.rs
```diff
@@ -0,0 +1,895 @@
...
+
+        Ok(self.take_payload_if_complete(spend.commit_txid))
+    }
+
+    fn insert_reveal(
+        entry: &mut IncompletePayloadEntry,
+        vout: u32,
+        envelope: SignedEnvelopeLeaf,
+        block_height: L1Height,
+    ) -> Result<(), CommitRevealParseError> {
```
delbonis: Did you review this AI generated code?

crates/l1proto/envelope-fmt/src/parser/payload.rs
```diff
@@ -0,0 +1,895 @@
...
+}
+
+/// Output produced by [`PayloadParser`] after parsing a batch of transactions.
+#[derive(Debug)]
+pub struct PayloadParserOutput {
+    /// Recovered payloads.
+    pub payloads: Vec<RecoveredPayload>,
+
+    /// Rejected transactions.
+    pub rejected_txs: Vec<RejectedTx>,
```
delbonis: What scenario would we care about these?  If not, then we can just discard them when we reject them.  If so, then maybe we shouldn't call them "rejected"?

delbonis (changes requested):
Un-pub all of the fields.  Private struct fields are available in fns within the same module, so it's not needed as convenience there.

crates/l1proto/envelope-fmt/src/parser/payload.rs
krsnapaudel: @delbonis The zkVM guest code calls with a set. Afaiu, the commit tx and reveal txs don't have the same height. This is based on the requirement that commit has to confirm before reveals are broadcast.

crates/l1proto/envelope-fmt/src/parser/payload.rs
krsnapaudel: @delbonis Yes this a point I am not sure about. The issue here is we cannot return error as soon as we encounter. Not sure if we should log in the library or just ignore them. If we ignore them, a lot of error types are not needed.

crates/l1proto/envelope-fmt/src/parser/payload.rs
krsnapaudel: @delbonis Yes the refactor into helper was done by Codex. So I did review before and after. What is the concern?

crates/l1proto/envelope-fmt/src/parser/payload.rs
```diff
@@ -0,0 +1,936 @@
...
+            let tx = included.tx;
+            let txid = tx.compute_txid();
+            let tail = match read_anchor_marker(&self.config.magic, tx) {
+                Some(tail) => tail,
+                None => {
+                    // `SingleReveal` anchor must have a valid marker.
+                    rejected.push(RejectedTx {
+                        txid,
+                        error: CommitRevealParseError::MalformedAnchorMarker,
+                    });
```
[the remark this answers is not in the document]
krsnapaudel: There is no point calling this with the whole block.

crates/l1proto/envelope-fmt/src/parser/payload.rs
delbonis: "with a set", what do you mean?

I'm not sure what that has to do with my review comment.

It seems like you're over-specializing this for the specific usage within the zkvm crate.

crates/l1proto/envelope-fmt/src/parser/payload.rs
delbonis: I also asked you to try, for a time, to entirely stop using AI, because it will hopefully improve your code, even if it takes you longer.

The issue I'm pointing at here is probably the second most common type of comment I leave on your code, after the error string formatting which is really more of a stylistic thing, but this is actually misleading.

crates/l1proto/envelope-fmt/src/parser/payload.rs
delbonis: What is the purpose in retaining these transactions?  Start there.

crates/l1proto/envelope-fmt/src/parser/payload.rs
krsnapaudel: @delbonis Afaiu, you want one height to be passed with a batch of txs. What I mean is two things:
1. The zkVM code does not call for commit and reveals separately. It just passes a batch of txs that includes both when asking to extract payload.
2. Irrespective of that, should this batch api handle commit and reveals coming in one pass? If so, we need the block heights separately. If we want the caller to call for commit and reveals separately, then one height will do.

crates/l1proto/envelope-fmt/src/parser/payload.rs
krsnapaudel: @delbonis I understand the point about AI. I have taken the important point about your feedback. It's not about obeying orders here. It's about building understanding. Now, what is misleading about this method?

crates/l1proto/envelope-fmt/src/parser/payload.rs
krsnapaudel: @delbonis The only purpose I can see is for observability.

crates/l1proto/envelope-fmt/src/parser/payload.rs
delbonis: This is not an instance fn or a constructor, and it's not part of a trait impl, so it should not be in this impl block.

crates/l1proto/envelope-fmt/src/parser/payload.rs
krsnapaudel: @delbonis So I realized zkVM guest code iterates over blocks. So this is not an issue. I will make the api accept one block height. Basically commit and reveals will go in separate batches because they will be in separate blocks.

irnb (approved):

PR #170, opened 2026-08-17
fix(service): wake a sync worker on shutdown instead of only on input

prajwolrg:
## Description

A sync service worker could ignore a shutdown signal indefinitely.

The loop in `sync_worker::worker_task` only checks its shutdown guard *after* `recv_next()` returns:

```rust
while let Some(input) = inp.recv_next()? {
    if shutdown_guard.should_shutdown() { break; }
    ...
```

`recv_next()` on a command channel calls `blocking_recv()`, which parks until the next message. The channel's senders are held by a `CommandHandle` that normally lives as long as the process, so the channel never closes on its own. A worker that is idle when shutdown fires therefore stays parked until the next message happens to arrive — which for a quiet service may be never. The process then hangs until the task manager's graceful-shutdown timeout expires and it is torn down the hard way.

The async worker does not have this problem: it already `select!`s the whole loop against `wait_for_shutdown()`. This brings the sync worker to parity. A sync worker runs on a plain thread that nothing can cancel from outside, so the wait itself has to be shutdown-aware.

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [x] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Notes to Reviewers

Found while fixing a related shutdown bug downstream in `asm` (alpenlabs/asm#229), where the ASM worker is a sync service driven by a command channel. Independent of that fix — it is why a fatal ASM worker exit was slow to take the process down even once it was raised properly. This PR is also independent of #169; they touch different files and can land in either order.

**The design choice worth your attention.** `SyncServiceInput` gains `recv_next_until_shutdown`, with a default that delegates to the existing `recv_next`. That keeps the change non-breaking for input implementations and is correct for inputs that cannot block indefinitely — `VecInput` and `IterInput` are finite and in-memory. The two adapters that *can* park forever, `TokioMpscInput` and `SyncAsyncInput`, override it. The downside of a permissive default is that a future blocking input silently inherits the old behaviour; the alternative was a required method, which breaks every implementor for no gain in the four in-tree cases. I went with the default and documented the expectation on the trait method. Say the word if you would rather have it required.

**Why `futures::executor::block_on` in the mpsc adapter** rather than a tokio runtime: both halves of the race are purely waker-driven — a tokio mpsc channel on one side, and on the other `strata-tasks`' `ShutdownGuard`, which is an `AtomicBool` plus a `Notify`. Neither needs the timer or the IO driver. A sync worker gets a plain `std::thread` from `spawn_critical` with no runtime entered, so there may not be one to borrow. `SyncAsyncInput` already holds an explicit `Handle` and keeps using it.

I used a biased select so a shutdown already in flight beats a queued message. The loop discarded that message anyway on the next guard check, so this only makes the exit prompter — no behaviour change.

**The breaking part.** `SyncExecutor::ShutdownGuard` now additionally requires `AsyncGuard + Sync`. This only breaks an out-of-tree `SyncExecutor` whose guard cannot await a shutdown. `strata_tasks::ShutdownGuard`, the only in-tree impl, already satisfies both.

**On the test.** `worker_task_exits_while_parked_on_idle_input` holds the sender open for the whole test, so nothing but the shutdown signal can end the wait, and it sleeps first to make sure the worker is genuinely parked rather than catching the pre-loop check. I confirmed it fails against the old loop — reverting just the `recv_next_until_shutdown` call makes it fail with "worker did not wake on the shutdown signal" after its 5s deadline, and pass again with the fix.

## Checklist

Full workspace `cargo test`, `cargo clippy --all-targets` and `cargo fmt --check` all pass.

## Related Issues

Unblocks prompt shutdown for alpenlabs/asm#229.

### DIFF
--- a/crates/service/src/adapters/mpsc.rs
+++ b/crates/service/src/adapters/mpsc.rs
@@ -1,8 +1,9 @@
 use std::future::Future;
 
+use futures::FutureExt;
 use tokio::sync::mpsc;
 
-use crate::{AsyncServiceInput, ServiceInput, ServiceMsg, SyncServiceInput};
+use crate::{AsyncGuard, AsyncServiceInput, ServiceInput, ServiceMsg, SyncServiceInput};
 
 /// Adapter for using a mpsc receiver as a input.
 ///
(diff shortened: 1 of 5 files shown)


crates/service/src/adapters/mpsc.rs
```diff
@@ -53,6 +54,46 @@ impl<T: ServiceMsg> SyncServiceInput for TokioMpscInput<T> {
...
+            futures::select_biased! {
+                // Biased so a shutdown already in flight wins over a queued
+                // message.  The loop below discards that message anyway once it
+                // sees the guard, so this only makes the exit prompter.
+                _ = shutdown.wait_for_shutdown().fuse() => None,
+                item = self.rx.recv().fuse() => Some(item),
+            }
+        };
+
+        match futures::executor::block_on(recv) {
```
delbonis, line 86: Will this work if the sync worker isn't run in a context with a runtime on its thread?

delbonis (changes requested):
This is very cleanly implemented but I'm not sure if I'm a big fan of this approach.  It makes the sync input concept more complex and it muddies the purpose of the base `recv_next` fn, and makes it less of a clean translation of the async version.  But it's possible that's just not avoidable (without actually going and doing weird stuff using unix signals).

Could this be implemented as another wrapper?  It might not be.

delbonis:
It's possible that a better way to implement this would involve having the shutdown handles hold a [`WeakSender`](https://docs.rs/tokio/latest/tokio/sync/mpsc/struct.WeakSender.html) (so that they don't hold them open when they'd naturally shut down) for these command services so that they can order a shutdown with a special message type.  Would require some work with the builder to make this more ergonomic.

PR #171, opened 2026-08-18
STR-4220: Schema versioning system

delbonis:
## Description

This PR defines a bunch of traits and supplementary types for a new schema versioning system.  With this new design, the database layer just stores opaque blobs and higher layer types which have more context for when/how migrations should be applied can do it on an as-needed basis.  This is unlike how migrations usually work in SQL databases, where the migration is done in a stop-the-world basis.  In some cases it's also undesirable to ever try to migrate old values to new versions (like blocks, state, etc.).

### Type of Change

- [ ] Documentation update
- [ ] Refactor
- [ ] New or updated tests
- [ ] Dependency update
- [ ] Security fix

## Checklist

### DIFF
--- a/crates/db-schema-common/src/containers.rs
+++ b/crates/db-schema-common/src/containers.rs
@@ -1,0 +1,84 @@
+use serde::{Deserialize, Serialize};
+
+use crate::types::*;
+
+/// Represents some container that has a well-defined version and a payload,
+/// which may store anything.
+///
+/// What kind of value it stores depends on the context and is assumed to be known.
+pub trait ValueContainer {
+    /// Gets the version this container has.
+    fn version(&self) -> VersionId;
+
+    /// Gets the encoded data.
+    fn payload(&self) -> &[u8];
+}
+
+/// Opaque container for a versioned value.
+///
+/// This stores the version it was stored as as a well-defined value, so that
+/// the higher machinery knows how to interpret the data.
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+pub struct OwnedValueContainer {
+    ver: VersionId,
+    pl: Vec<u8>,
+}
+
+impl OwnedValueContainer {
+    /// Creates a new instance.
+    pub fn new(ver: VersionId, pl: Vec<u8>) -> Self {
+        Self { ver, pl }
+    }
+}
+
+impl ValueContainer for OwnedValueContainer {
+    fn version(&self) -> VersionId {
+        self.ver
+    }
+
+    fn payload(&self) -> &[u8] {
+        &self.pl
+    }
+}
+
+/// Borrowed version of [`ValueContainer`] to reduce copies during the first
+/// deserialization passes before we deserialize it to a concrete value.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Deserialize)]
+pub struct ValueContainerRef<'b> {
+    ver: VersionId,
+    pl: &'b [u8],
+}
+
+impl<'b> ValueContainerRef<'b> {
+    /// Creates a new instance.
+    pub fn new(ver: VersionId, pl: &'b [u8]) -> Self {
+        Self { ver, pl }
+    }
+}
+
+impl<'b> ValueContainer for ValueContainerRef<'b> {
+    fn version(&self) -> VersionId {
+        self.ver
+    }
+
+    fn payload(&self) -> &[u8] {
+        self.pl
+    }
+}
+
+/// Extension trait for [`ValueContainer`] to provide convenience fns.
+pub trait ValueContainerExt {
+    /// Attempts to decode the payload as a particular schema version.
+    ///
+    /// # Panics
+    ///
+    /// If the version does not match.
+    fn try_decode_as_ver<S: Schema, V: SchemaVersion<S>>(&self) -> Result<V, S::Error>;
+}
+
+impl<T: ValueContainer> ValueContainerExt for T {
+    fn try_decode_as_ver<S: Schema, V: SchemaVersion<S>>(&self) -> Result<V, S::Error> {
+        assert_eq!(V::VERSION, self.version(), "schema: version mismatch");
+        unimplemented!()
+    }
+}
--- a/crates/db-schema-common/src/containers.rs
+++ b/crates/db-schema-common/src/containers.rs
@@ -1,5 +1,8 @@
+//! Containers for versioned, opaquely-encoded values.
+
 use serde::{Deserialize, Serialize};
 
+use crate::errors::DecodeValueError;
 use crate::types::*;
 
 /// Represents some container that has a well-defined version and a payload,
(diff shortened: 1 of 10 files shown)


crates/db-schema-common/src/types.rs
```diff
@@ -0,0 +1,70 @@
...
+/// }
+/// ```
+pub trait SchemaVersion<S: Schema> {
+    /// The version of the schema this type embodies.
+    const VERSION: VersionId;
+}
+
+/// Defines a type that contains variants for each possible schema version.
+// TODO(trey): a lot of this is speculative
+pub trait SchemaEnum<S: Schema> {
```
storopoli: What if we call this `SchemaSeries`? I know we started to bike shed this in STR-4220, but we can resume bike shedding here.

crates/db-schema-common/src/types.rs
```diff
@@ -0,0 +1,70 @@
...
+    /// Exposes the newest type.
+    type Highest: SchemaVersion<S>;
+
+    /// The ID of the version this variant currently is.
+    fn version(&self) -> VersionId;
+
+    /// If possible, upgrades to the next-newest version.
+    ///
+    /// If this is already the highest version, this MUST return `self` unchanged.
+    fn upgrade(self) -> Self;
```
storopoli: What if we have a `upgrade_to_highest` in this trait as well?

crates/db-schema-common/src/types.rs
```diff
@@ -0,0 +1,70 @@
...
+///
+/// ```
+/// struct Foo;
+///
+/// impl Schema for Foo {
+///     const KEY: &str = "foo";
+/// }
+/// ```
+pub trait Schema {
+    /// Universal key type.
```
storopoli: Bikeshed: This is more of a "value" than a "type"

```suggestion
    /// Universal key value.
```

crates/db-schema-common/src/containers.rs
```diff
@@ -0,0 +1,84 @@
...
+
+/// Opaque container for a versioned value.
+///
+/// This stores the version it was stored as as a well-defined value, so that
+/// the higher machinery knows how to interpret the data.
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize, Serialize)]
+pub struct OwnedValueContainer {
+    ver: VersionId,
+    pl: Vec<u8>,
+}
```
storopoli, line 29: With a derived `Serialize`, `pl: Vec<u8>` encodes as a CBOR array of individual integers, not a byte string — roughly 2x size for every stored blob. Maybe use `serde_bytes` (or ciborium's byte-string handling) on the payload `pl` field.

crates/db-schema-common/src/containers.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+}
+
+/// Borrowed version of [`ValueContainer`] to reduce copies during the first
+/// deserialization passes before we deserialize it to a concrete value.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Deserialize)]
+pub struct ValueContainerRef<'b> {
+    ver: VersionId,
+    pl: &'b [u8],
+}
```
storopoli: `ValueContainerRef<'b>` derives `Deserialize` around `&'b [u8]`, but ciborium does not support borrowed byte-slice deserialization — that derive will fail at runtime for the format it's presumably meant for, defeating the zero-copy purpose. A round-trip test would have caught both...

storopoli (commented):
Very minor suggestions. Overall this is in the right track!

crates/db-schema-common/src/containers.rs
delbonis: Crap!  I keep forgetting this!

crates/db-schema-common/src/types.rs
delbonis: That's a better name.

crates/db-schema-common/src/types.rs
delbonis: I'm actually gonna remove this probably since the migrations aren't really inherent to the types.

crates/db-schema-common/src/containers.rs
delbonis: Maybe I should use flexbuffers...

storopoli: No, let's not add more dependencies. I think we can work around with ciborium somehow.

delbonis: figured it out

crates/db-schema-common/src/migrator.rs
```diff
@@ -0,0 +1,396 @@
...
+
+        // This also bounds how far a chain walk can run, since it can only
+        // proceed through versions we have an entry for.
+        assert!(
+            B::VERSION <= MAX_VERSION_ID,
+            "schema/migrator: version {} is above the max version {MAX_VERSION_ID}",
+            B::VERSION
+        );
+
+        let sch_tbl = self.tbl.entry(S::KEY).or_default();
```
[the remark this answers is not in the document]
storopoli: I think this is valid @delbonis.
Fix: key on `TypeId::of::<S>()` (or `(TypeId, KEY)`), which also removes string hashing from the read path.

crates/db-schema-common/src/migrator.rs
```diff
@@ -0,0 +1,396 @@
...
+use crate::errors::MigrationError;
+use crate::types::*;
+
+/// Opaque migration trait that we put behind a `Box<dyn Migration>` to hide the
+/// schema version-specific generics in `FnMigration`.
+///
+/// The three steps are separated so that a chain of migrations can decode once
+/// at the head and encode once at the tail, applying each intermediate step in
+/// value space instead of round-tripping through bytes at every version.
+trait Migration: 'static {
```
storopoli:
```suggestion
trait Migration: Send + Sync + 'static {
```

This needs to be `Send` and `Sync`.

A migration registry needs to sit behind `Arc`/`OnceLock` and be read from many threads.

crates/db-schema-common/src/containers.rs
storopoli: This is still unresolved. Gently bump.

storopoli (changes requested):
Couple correction stuff. Once we fix this we can merge.

prajwolrg:
We are building similar thing on the EE side in alpenlabs/alpen-ee#114 — per-value version tag, lazy migration on read, frozen version structs. The parts that differ might be useful here

The main one: the migration chain is types, not a runtime registry. `versioned_value! { 1 => FooV1, 2 => FooV2 }` generates the fold to current by calling `UpConvert<Next>` on each consecutive pair, so a missing `N -> N+1` hop is a build error and there's no `Box<dyn Any>` on the path. `ChainTypeMismatch` stops being a possible failure at all, and the schema-key collision and `Send + Sync` points raised above go away with it, since there's no shared registry left to key or to share.

Two smaller ones:

- Up-converters get a context that can read other tables in the ambient transaction, so a new field can be derived from other state instead of only defaulted. Depth-capped, so a converter read cycle is an error rather than a stack overflow.
- Golden fixtures: every declared version keeps an archived sample of its on-disk bytes, and CI replays them through the current decode path. Old converters only ever run on old data, so without this they rot quietly.

The design is described [here](https://app.notion.com/p/ee-schema-upgrades-39d901ba000f8001a4fad1371c1c9c6a).

delbonis:
@prajwolrg Do you think it would make sense to just upstream the common libraries here and drop this whole PR then?

storopoli:
> @prajwolrg Do you think it would make sense to just upstream the common libraries here and drop this whole PR then?

I'm slightly in favor of this.

prajwolrg:
> @prajwolrg Do you think it would make sense to just upstream the common libraries here and drop this whole PR then?

That can work. The PR on alpen-ee isn't merged yet and is still going through the review, although I don't see a major blocker. If the design fulfills the requirements for the OL well, it's better reuse what we’ve already built here rather than maintain two implementations. 

Feel free to use the existing work to update this PR or open a new PR entirely.

Once the PR lands and is also used in strata-ol, we can then update the alpen-ee implementation to use the upstream crates.

storopoli:
@evgenyzdanovich since @delbonis is off for the whole week I took the liberty of addressing several correctness issues that I've pointed out in previous reviews.

Please take a look and review this such that we can get this merged.

storopoli (approved):
ACK 62e9401a1f17f1e5bfb6467de77e1314d2668f3f

delbonis:
Changes look good.  I would like to consider the possibility of borrowing the alpen-ee migrator system but if @prajwolrg is okay with this as-is then we can figure out the specifics later.

